PR #1, opened 2025-11-20
patch: [DPE-8871]: OpenSearch - initialise single kernel workflow

akram09:
This Pull Request provides a starting point for the refactoring of OpenSearch charm to a single-kernel architecture. It also provide a starting point for the new kubernetes charm for OpenSearch.

The content of the PR includes mainly two main important things:
1. Single Kernel Configuration: This refers to the set of configuration files that were added to support development and testing of the library including pyproject.toml, poetry.lock, tox.ini, and pre-commit. I have used mostly the same configurations that were in the previous [opensearch-operator](https://github.com/canonical/opensearch-operator) repository.
2. For local development and testing, the charms under `tests/charms/` are example charms that uses single-kernel library. 

image

### Local Development 
To build the charm locally, we can leverage the `./scripts/build_lib_for_integration.sh` which will copy the library to the test charm and use it to build an opensearch charm using the single-kernel library.


.github/workflows/check_pr.yaml
```diff
@@ -0,0 +1,16 @@
...
+# See LICENSE file for licensing details.
+name: Check pull request
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+    branches:
+      - 2/edge
```
Mehdi-Bendriss, line 11: I would probably also add, for the time being, the current branch `feature/single-kernel` here

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+name: Release to PyPI
+
+on:
+  push:
+    branches:
+      - 6/edge
```
Mehdi-Bendriss:
```suggestion
      - 2/edge
```

.github/workflows/release.yaml
Mehdi-Bendriss, line 1: i think we need a second release pipeline that gets triggered for the `feature/single-kernel` branch, for [testpypi](https://test.pypi.org/)?

.pre-commit-config.yaml
```diff
@@ -48,10 +48,6 @@ repos:
     hooks:
     - id: isort
 
-  - repo: https://github.com/pycqa/flake8
```
Mehdi-Bendriss, line 51: why not?

Mehdi-Bendriss (commented):
Thanks Akram! Looks good. Only a few comments

Mehdi-Bendriss (approved):
Thanks Akram!

smethnani (approved):
Thank you @akram09!

PR #5, opened 2026-01-06
[patch]: TLS Management Refactoring

akram09:
This Pull Request propose a new refactoring of TLS management in OpenSearch. It aims to provide a new way of handling events leveraging the new event/manager/state architecture. Basically, the main changes are the following:
- Adding a TLSEventsHandler which receives and handle the events related to the charm integration with TLS providers. The main events that are handled are `on_tls_relation_created` `on_certificate_available` `on_set_private_key` `on_certificate_expiring` `on_certificate_invalidated`. 
- Adding a TLSManager which abstracts the TLS management business logic by interacting with both the workload and the charm state. The manager can also be used by other eventsHandler since it provides basic functionalities to interact with the keystore, truststore, and managing the certificates. 

The new workflow for both `on_tls_relation_created` and `on_certificate_available` are shown below:
image

image


opensearch_single_kernel/common/constants.py
s-mlnk, line 117: **praise** great change!

opensearch_single_kernel/core/state.py
```diff
@@ -50,6 +50,8 @@
 if TYPE_CHECKING:
     from opensearch_single_kernel.charms.base import OpenSearchBaseCharm
 
+logger = logging.getLogger(__name__)
+
```
s-mlnk: **todo** it's a duplicate.

opensearch_single_kernel/core/state.py
```diff
@@ -358,6 +409,131 @@ def get_unit(self, name: str):
...
+            if (
+                self.peer_relation.data[unit].get("tls_configured") != "True"
+                or "tls_ca_renewing" in self.peer_relation.data[unit]
+                or "tls_ca_renewed" in self.peer_relation.data[unit]
+            ):
+                return False
+        return True
+
+    @property
+    def ca_rotation_complete_in_cluster(self) -> bool:
```
s-mlnk, line 448: **suggestion** it will be more convenient to go through a OpenSearchServer list generator for all the units just like:
```python
class ClusterState(Object):

    @property
    def servers(self) -> Iterable[OpenSearchServer]:
        return [
            OpenSearchServer(
                relation=self.peer_relation,
                data_interface=self.peer_unit_interface,
                component=unit,
            )
            for unit in self.all_units
        ]

    @property
    def ca_rotation_complete_in_cluster(self) -> bool:
        """Check whether the CA rotation completed in all units."""

        return not any([server.tls_ca_renewing for server in self.servers]) or all(
            [server.tls_ca_renewed for server in self.servers]
        )
```
Same with `ca_and_certs_rotation_complete_in_cluster`.

opensearch_single_kernel/core/state.py
s-mlnk, line 356: **todo** it's mixed up. It should be `peer_cluster_orchestrator with PEER_CLUSTER_ORCHESTRATOR_RELATION` and `peer_cluster with PEER_CLUSTER_RELATION`.

s-mlnk: **suggestion** if it's a static value, it should be moved to constants.

opensearch_single_kernel/managers/external_clients.py
s-mlnk, line 1: **question** what purpose of this manager? I don't see any references.

opensearch_single_kernel/workload/base.py
```diff
@@ -119,6 +120,12 @@ def seed_hosts(self) -> PathProtocol:
 class BaseWorkload(ABC):
     """Base interface for common workload operations."""
 
+    @property
+    @abstractmethod
+    def root(self) -> PathProtocol:
+        """Return the root path."""
+        pass
+
```
s-mlnk, line 130: **question** why exactly do you need root? This question continues [thread on `write_file` function](https://github.com/canonical/opensearch-single-kernel-library/pull/5/changes#r2728130243).

opensearch_single_kernel/workload/base.py
```diff
@@ -130,6 +137,42 @@ def paths(self) -> Paths:
...
+        """Persists data into file. Useful for files generated on the fly, such as certs etc."""
+        if not override and self.exists(path_str):
+            return
+        path = self.root / path_str
+
+        parent_dir_path = path.parent
+        if parent_dir_path:
+            parent_dir_path.mkdir(parents=True, exist_ok=True)
+
+        path.write_text(data)
```
s-mlnk: 1. I don't see any usages of `override`, so why do we need that functionality?
2. I only see a single usage of this function, in which you use PathOps (`managers/tls.py::update_request_ca_bundle`). Why do we need to convert path to str?
3. Same applies to `exists` function, only usage of which is here.

opensearch_single_kernel/workload/base.py
```diff
@@ -130,6 +137,42 @@ def paths(self) -> Paths:
...
+
+    def exists(self, path_str: str) -> bool:
+        """Return whether the path exists in filesystem."""
+        path = self.root / path_str
+        return path.exists()
+
+    @abstractmethod
+    def remove_file(self, file_path: str):
+        """Remove file from the filesystem."""
+        pass
```
s-mlnk: **suggestion** if there are no usages of this function, it can be safely removed to minimise codebase.

opensearch_single_kernel/workload/base.py
```diff
@@ -130,6 +137,42 @@ def paths(self) -> Paths:
...
+        path = self.root / path_str
+
+        parent_dir_path = path.parent
+        if parent_dir_path:
+            parent_dir_path.mkdir(parents=True, exist_ok=True)
+
+        path.write_text(data)
+
+    @contextmanager
+    def temp_file(
```
s-mlnk, line 203: **question** considering all of the usages, may this logic be simplified by adding `body` parameter to this function? So, instead of 
```
with workload.temp_file(...) as file:
    file.write("my test data")
    file.flush()
```
it may be
```
filepath: PathOps = workload.write_temp_file(..., body="my test data")
```
This way, it will simplify logic later on.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,440 @@
...
+            self.certs.request_certificate_creation(certificate_signing_request=unit_http_csr)
+
+    def _on_tls_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Notify the charm that the relation is broken."""
+        # TODO: If upgrade log a warning
+        if self.charm.tls_manager.all_tls_resources_stored():
+            return
+
+        # Otherwise, we block.
+        self.status.set(CharmStatuses.TLS_RELATION_BROKEN)
```
s-mlnk: **todo** unknown attribute `self.status`.

tests/conftest.py
```diff
@@ -13,9 +13,9 @@ def pytest_addoption(parser: Parser):
     parser.addoption(
         "--substrate",
         action="store",
-        help="Substrate to test, either lxd or microk8s",
-        choices=("lxd", "microk8s"),
-        default="lxd",
```
s-mlnk, line 18: **todo** `tests/helpers.py::Substrate` also should be changed.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,440 @@
...
+        self.charm.tls_manager.store_admin_tls_secrets_if_applies()
+
+        # In case of renewal of the unit transport layer cert - restart opensearch
+        if renewal and self.charm.state.application.is_admin_user_initialized:
+            if self.charm.tls_manager.is_fully_configured():
+                try:
+                    self.charm.tls_manager.reload_tls_certificates()
+                except OpenSearchHttpError:
+                    logger.error("Could not reload TLS certificates via API, will restart.")
+                    self.charm._restart_opensearch_event.emit()
```
reneradoi: This event should be renamed if accessed from outside of the charm module itself.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -477,6 +572,70 @@ def _post_start_init(self, event: StartOpenSearch) -> None:
         self.charm.status.clear(CharmStatuses.SERVICE_START_ERROR)
         self.charm.status.clear(CharmStatuses.PEER_CLUSTER_NO_DATA_NODE)
 
+        # TODO: Handle event.after_upgrade
+        # TODO: Handle refresh relation data of peer cluster
+
+        self.charm.tls_events.configure_tls_after_start()
```
reneradoi: We should avoid calling one event handler from another event handler.

This is another example of accidential architecture (also see comment to `tls_events.request_new_unit_certificates`): Because it calls a method of the TLS requirer class, it needs to be implemented in the tls event handler. As this is directly related to the (post-) start of the charm, you would rather want to have the status handling itself here in this method, and call the TLS manager from here. This will be possible as soon as `request_new_unit_certificates` was moved to the TLS manager.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,440 @@
...
+            self.charm.tls_manager.finalize_ca_certs_rotation()
+
+        # TODO: Handle case of peer cluster manager
+        # if self.peers_data.get(Scope.UNIT, "cluster_manager_removed", default=False):
+        # restore cluster_manager role and restart the service
+        # logger.debug("Restoring cluster_manager role and restarting the service")
+        # self.peers_data.delete(Scope.UNIT, "cluster_manager_removed")
+        # self._restart_opensearch_event.emit()
+
+    def request_new_unit_certificates(self) -> None:
```
reneradoi: This method (as well as for the admin cert) should rather be part of the manager.

I understand that it is tempting to have these as part of the event handler, given they call `self.certs.request_certificate_revocation` which itself depends on the `TLSCertificatesRequiresV3` being instantiated in the same class. Nevertheless, this leads to a few methods being implemented in the event handler that should actually be in the manager class instead, creating "accidential architecture". Also the separation of responsibilities between event handler and manager become very blurred because of this, for example you have `create_certificate_signing_request` on the manager while having `request_new_unit/admin_certificates` on the event handler.

You can solve this for example with a custom event. This event could be called from the manager and be handled here, to then call `self.certs.request_certificate_revocation`. Thereby you can separate concerns a lot better, and as a side effect, won't have to call `tls_events.request_new_unit_certificates` from other event handlers (which will probably happen at some point).

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,779 @@
...
+            # ):
+
+            self.state.secrets.put_object(
+                scope,
+                cert_type.val,
+                {f"{store_type.val}-password": generate_password()},
+                merge=True,
+            )
+
+    def _get_subject(self, cert_type: CertType) -> str:
```
reneradoi, line 152: This could be a property, as it doesn't require computation and is rather static.

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,779 @@
...
+                            "Failed to import cert for alias %s into %s: %s",
+                            internal_alias,
+                            store_path,
+                            (e.out or "") + (e.err or ""),
+                        )
+                        return False
+            except OSError as e:
+                # tmp file creation issues
+                logger.error("Failed to create temporary file for CA import: %s", e)
+                return False
```
reneradoi: I've noticed that you have moved a lot of code to the manager that I would rather consider responsibility of the workload. This block is a good example, and it seems almost anti-pattern to do so. There are more examples in this class, for instance `add_ca_to_request_bundle` or `store_key_pair`. If you look at them again you will realise that they go against the layers of abstraction we actually want to have, where a manager bundles business logic while the workload does all the low-level stuff such as file handling. 

As one obvious consequence, you have a lot of the `temp_file` handling repeating in the different methods of the manager class. But more importantly, you create a close coupling between a (rather abstract) manager and the low-level implementation for the `PathProtocol` -> whenever this changes, you will have to adjust all the manager code that uses it.

Rule of thumb should be (I guess): whenever you need to import `PathProtocol` to the manager class, think again if this should really happen there.

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,779 @@
...
+    def _remove_ca_from_request_bundle(self, ca_cert: str) -> None:
+        """Remove the CA cert from the request bundle for the requests module."""
+        bundle_path = self.workload.paths.certs / "chain.pem"
+        if not bundle_path.exists():
+            return
+
+        bundle_content = bundle_path.read_text()
+        bundle_path.write_text(bundle_content.replace(ca_cert, ""))
+
+    def store_new_ca(self, secrets: dict[str, Any], create_store_pwd: bool) -> bool:  # noqa: C901
```
reneradoi: I think the `noqa` might not be required here anymore.

reneradoi (commented):
Hey Akram! Again I have focused my review more on the conceptual design than on implementation details. I would recommend to have another look at separating concerns between TLS events, TLS manager and workload. Please refer to the comments for more details. You can also see this on the sequence diagrams (btw. I really like how you do this, great job!): when an event handler calls a method on itself, this should only happen to avoid duplicated code. As soon as "business logic" happens in an event handler method, it should move out of the event handler (for instance: request new certificate). You've done this really good with `store_new_ca`, which on the diagram is still on the event handler, but the implementation has already been moved to the manager. Keep doing this!

opensearch_single_kernel/managers/tls.py
s-mlnk: I agree with point that `tmp.write()` & `tmp.flush()` & `tmp.name` should be internal to the workload, but there isn't any `PathProtocol` usage here. Actually, `temp_file` returns a raw file object, that is much worse. And even if there was `PathProtocol` - it would be great instead, because it's an interface for universal file handling by `LocalPath` & `ContainerPath`, not the "low-level stuff". So, in my opinion, usage of `PathProtocol` should not be prohibited but encouraged)

opensearch_single_kernel/managers/external_clients.py
akram09: This manager will handle the opensearch-provider relation which will be added in a future PR

opensearch_single_kernel/workload/base.py
akram09: We would need the root in order to create the paths using pathops. This is stated in their documentation

opensearch_single_kernel/workload/base.py
akram09: I think override will be used by `snapshots` PR as I recall from the previous opensearch-operator repository

akram09: I deleted the function thank you.

opensearch_single_kernel/workload/base.py
akram09: I personally prefer to leave it as a context manager to give more flexibility of usage

opensearch_single_kernel/events/opensearch.py
akram09: Thank you Rene

opensearch_single_kernel/workload/base.py
akram09: We would need a context manager to clean the file later. I will modify to return a PathProtocol instead.

opensearch_single_kernel/common/statuses.py
```diff
@@ -20,6 +20,11 @@ class CharmStatuses(Enum):
     # TLS Status
     TLS_RELATION_MISSING = BlockedStatus("Missing TLS relation with this cluster.")
     TLS_NOT_FULLY_CONFIGURED = MaintenanceStatus("Waiting for TLS to be fully configured...")
+    TLS_CA_ROTATION = MaintenanceStatus("Applying new CA certificate...")
+    TLS_RELATION_BROKEN = BlockedStatus(
+        "Relation broken with the TLS Operator while TLS not fully configured. Stopping OpenSearch."
+    )
+    TLS_CERTS_EXPIRATION_ERROR = BlockedStatus("The certificates:  need to be refreshed.")
```
gatici:
```suggestion
    TLS_CERTS_EXPIRATION_ERROR = BlockedStatus("The certificates: {} need to be refreshed.")
```

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+        if scope == Scope.UNIT:
+            admin_secrets = (
+                self.charm.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True)
+                or {}
+            )
+            if not (truststore_pwd := admin_secrets.get("truststore-password")):
+                event.defer()
+                return
+
+            keystore_pwd = self.charm.state.secrets.get_object(scope, cert_type.val, peek=True)[
```
gatici, line 339: Are we sure that keystore_pwd always exist ?  Should we defer here if unit_secrets does not exist:
```
 unit_secrets = self.charm.state.secrets.get_object(scope, cert_type.val, peek=True)
            if not unit_secrets or "keystore-password" not in unit_secrets:
                event.defer()
                return
```

opensearch_single_kernel/managers/config.py
```diff
@@ -219,6 +221,58 @@ def add_seed_hosts(self, cm_ips: list[str]):
             lines = "\n".join([entry for entry in cm_ips_set if entry.strip()])
             self.workload.paths.seed_hosts.write_text(f"{lines}\n")
 
+    def set_admin_tls_conf(self, secrets: dict[str, any]):
```
gatici: nit:
```suggestion
    def set_admin_tls_conf(self, secrets: dict[str, Any]):
```

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,767 @@
...
+        # post-actions
+        try:
+            command = ""
+            if snap_user_with_write_permission:
+                command = (
+                    f"sudo chown {snap_user} {store_path}; sudo chmod {final_mode} {store_path};"
+                )
+            if add_read_perm:
+                command += f"sudo chmod +r {store_path}"
+            self.workload.run_cmd(command)
```
gatici: We can check if command exist or not before directly running it.

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,767 @@
...
+        ) as tmp_ca_file:
+
+            try:
+                return self.workload.run_cmd(f"openssl x509 -in {tmp_ca_file} -noout -issuer").out
+            except OpenSearchCmdError as e:
+                logger.error("Error reading the current truststore: %s", e)
+                return None
+
+    def get_cert_issuer_from_path(self, store_pwd: str, store_path: PathProtocol) -> str | None:
+        """Retrieve the certificate issuer from a string certificate."""
```
gatici:
```suggestion
        """Retrieve the certificate issuer from the cert in the given PKCS12 store.."""
```

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,767 @@
...
+        for cert_type in list(certs.keys()):
+            hours = cert_expiration_remaining_hours(certs[cert_type])
+            if hours > 24 * 7:
+                del certs[cert_type]
+
+        return certs
+
+    def remove_old_ca(self) -> None:
+        """Remove old CA cert from trust store."""
+        secrets = self.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True)
```
gatici, line 517: The `get_object` may return None. It's better to check secrets availability before processing.

opensearch_single_kernel/utils/helpers.py
```diff
@@ -99,3 +106,50 @@ def deployment_type(
...
+
+
+def normalized_tls_subject(subject: string) -> str:
+    """Removes any / character from a subject."""
+    if subject.startswith("/"):
+        subject = subject[1:]
+    return subject.replace("/", ",")
+
+
+def cert_expiration_remaining_hours(cert: string) -> int:
```
gatici:
```suggestion
def cert_expiration_remaining_hours(cert: str) -> int:
```

opensearch_single_kernel/utils/helpers.py
```diff
@@ -99,3 +106,50 @@ def deployment_type(
...
+
+
+def split_ca_chain(pem_content: str) -> list[str]:
+    """Split PEM chain into individual certificates."""
+    end_cert_marker = "-----END CERTIFICATE-----"
+    parts = [part.strip() for part in pem_content.split(end_cert_marker) if part.strip()]
+    return [f"{part}\n{end_cert_marker}" for part in parts]
+
+
+def normalized_tls_subject(subject: string) -> str:
```
gatici:
```suggestion
def normalized_tls_subject(subject: str) -> str:
```

opensearch_single_kernel/workload/vm.py
```diff
@@ -88,6 +122,14 @@ def get_host_public_ip(self) -> str | None:
 
         return output.out.strip()
 
+    def is_started(self) -> bool:
+        """Check if OpenSearch is started."""
+        reachable = self.is_reachable(self.host, self.port)
```
gatici: self.host and self.port is not defined.

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1672 @@
...
+    )
+    mocker.patch(
+        "opensearch_single_kernel.lib.charms.tls_certificates_interface.v3.tls_certificates.TLSCertificatesRequiresV3.request_certificate_creation"
+    )
+    event_mock = MagicMock()
+
+    harness.set_leader(is_leader=True)
+    harness.charm.tls_events._on_tls_relation_created(event_mock)
+
+    create_certificate_signing_request.mock_calls == [
```
gatici: we need to add assertion here:
```suggestion
    assert create_certificate_signing_request.mock_calls == [
```

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1672 @@
...
+    truststore_password = "12345"
+    harness.charm.state.secrets.put_object(
+        Scope.APP,
+        CertType.APP_ADMIN.val,
+        {"truststore-password": truststore_password},
+    )
+
+    harness.set_leader(is_leader=False)
+    harness.charm.tls_events._on_tls_relation_created(event_mock)
+    create_certificate_signing_request == [
```
gatici: We need to add assertion:
```suggestion
    assert create_certificate_signing_request == [
```

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,767 @@
...
+
+        start_cert_marker = "-----BEGIN CERTIFICATE-----"
+        chains: dict[str, list[tuple[int, str]]] = {}
+
+        for block in cert_blocks:
+            # find the friendlyName: line produced by openssl pkcs12
+            alias_line = next(
+                (line for line in block.split("\n") if line.strip().startswith("friendlyName:")),
+                None,
+            )
```
gatici, line 315: alias_line can be None. 
It is better to add continue here if alias_line is None.

gatici (commented):
Great work @akram09 ! 
I have some comments but looks good.

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,767 @@
...
+        if is_secret_found(u_http_secrets):
+            return Scope.UNIT, CertType.UNIT_HTTP, u_http_secrets
+
+        return None
+
+    def read_ca(self, alias: str, store_pwd: str, store_path: PathProtocol) -> str | None:
+        """Load stored CA cert."""
+        return (self.list_cas(store_pwd, store_path) or {}).get(alias)
+
+    def list_cas(
```
gatici: If these helper methods are useful to multiple managers, would it make sense to move them into a utils module (or a shared helper) instead of attaching them to a single manager class? That would avoid coupling and make reuse clearer.

opensearch_single_kernel/common/statuses.py
akram09: We don't need this. since it will be filled automatically with a dynamic message

opensearch_single_kernel/events/tls.py
akram09: The `unit_secrets` will always be there since we are checking in the first step of `on_certificate_available` . Keystore are created on the `tls-joined` If the keystore is not created the csr will not be requested and this event won't be called. This is my understanding.

opensearch_single_kernel/workload/vm.py
akram09: Removed since it is not used.

opensearch_single_kernel/managers/tls.py
akram09: The thing is that some of these are related to the workload. My idea is that if another manager would need to use these methods then it would make more sense to me to add them to the base manager rather than having a util function with workload as a parameter.

opensearch_single_kernel/common/statuses.py
gatici: This looks like certificate names are needed. we can  remove the `:` and  space before need.

opensearch_single_kernel/managers/config.py
```diff
@@ -219,6 +222,58 @@ def add_seed_hosts(self, cm_ips: list[str]):
             lines = "\n".join([entry for entry in cm_ips_set if entry.strip()])
             self.workload.paths.seed_hosts.write_text(f"{lines}\n")
 
+    def set_admin_tls_conf(self, secrets: dict[str, Any]):
+        """Configures the admin certificate."""
+        self.yaml_setter.put(
+            self.CONFIG_YML,
+            "plugins.security.authcz.admin_dn/{}",
+            f"{normalized_tls_subject(secrets['subject'])}",
```
gatici: normalized_tls_subject(secrets['subject']) already returns string, no need to wrap it with an f-string.

opensearch_single_kernel/managers/config.py
```diff
@@ -219,6 +222,58 @@ def add_seed_hosts(self, cm_ips: list[str]):
...
+            self.yaml_setter.put(
+                self.CONFIG_YML,
+                f"plugins.security.ssl.{target_conf_layer}.{store_type}_type",
+                "PKCS12",
+            )
+
+            self.yaml_setter.put(
+                self.CONFIG_YML,
+                f"plugins.security.ssl.{target_conf_layer}.{store_type}_filepath",
+                f"{self.workload.paths.certs_relative}/{cert if cert == 'ca' else cert_type}.p12",
```
gatici:
```suggestion
                f"{self.workload.paths.certs_relative}/{cert if cert == 'ca' else cert_type.val}.p12",
```

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1670 @@
...
+    assert "/var/snap/opensearch/current/etc/opensearch" in str(
+        tempfile.call_args_list[0][1]["dir"]
+    )
+    # NOTE: The new cert and chain are NOT saved into the keystore (disk)
+
+    # Set flag, set status, restart
+    assert harness.charm.state.server.tls_ca_renewing
+    assert isinstance(harness.model.unit.status, MaintenanceStatus)
+    assert harness.model.unit.status.message == CharmStatuses.TLS_CA_ROTATION.value.message
+    assert harness.model.unit.status, MaintenanceStatus != original_status
```
gatici:
```suggestion
    assert harness.model.unit.status != original_status
```
we need to compare original status with a single status.

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1670 @@
...
+                old_certificate_signing_request=b"csr-transport",
+                new_certificate_signing_request=generate_csr(),
+            ),
+        ]
+    )
+
+    # new admin cert
+    assert request_certificate_creation.call_count == 1
+    # we store the decoded csr in the secret but pass it as bytes to the function
+    request_certificate_creation.call_args.kwargs[
```
gatici: missing assert here

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1670 @@
...
+        "certificate_signing_request"
+    ].decode() == new_app_admin_secret["csr"]
+
+    assert (
+        harness.model.unit.status.message == CharmStatuses.TLS_NOT_FULLY_CONFIGURED.value.message
+    )
+    assert harness.model.unit.status, MaintenanceStatus != original_status
```
gatici:
```suggestion
    assert harness.model.unit.status != original_status
```
this pattern does not look right to me.

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1670 @@
...
+    )
+    assert (
+        harness.charm.state.secrets.get_object(Scope.UNIT, CertType.UNIT_TRANSPORT.val)["csr"]
+        != csr_transport_old
+    )
+
+    assert (
+        harness.model.unit.status.message == CharmStatuses.TLS_NOT_FULLY_CONFIGURED.value.message
+    )
+    assert harness.model.unit.status, MaintenanceStatus != original_status
```
gatici:
```suggestion
    assert harness.model.unit.status != original_status
```

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1670 @@
...
+        "cert": cert,
+        "chain": chain[0],
+        "truststore-password": "truststore_12345",
+        "keystore-password": "keystore_12345",
+        "key": key,
+        "ca-cert": ca,
+    }
+
+    assert harness.model.unit.status.message == ""
+    assert harness.model.unit.status, MaintenanceStatus != original_status
```
gatici:
```suggestion
    assert harness.model.unit.status != original_status
```

tests/unit/test_tls_manager.py
```diff
@@ -0,0 +1,1670 @@
...
+    assert re.search("keytool .*-delete .*-alias old-ca", run_cmd.call_args_list[-1].args[0])
+    assert "/var/snap/opensearch/current/etc/opensearch" in str(
+        tempfile.call_args_list[0][1]["dir"]
+    )
+
+    assert not harness.charm.state.server.tls_ca_renewing
+    assert not harness.charm.state.server.tls_ca_renewed
+
+    assert harness.model.unit.status.message == ""
+    assert harness.model.unit.status, MaintenanceStatus != original_status
```
gatici:
```suggestion
    assert harness.model.unit.status != original_status
```

gatici (commented):
Thank you Akram.
I provided some minor comments.

opensearch_single_kernel/workload/base.py
```diff
@@ -130,6 +137,26 @@ def paths(self) -> Paths:
...
+        errors=None,
+        suffix=None,
+    ):
+        """Context manager for creating temporary files."""
+        pass
+
+    def exists(self, path_str: str) -> bool:
+        """Return whether the path exists in filesystem."""
+        path = self.root / path_str
+        return path.exists()
```
s-mlnk: Considering we have `path.exists()`, I don't think this function is useful here.

opensearch_single_kernel/core/state.py
```diff
@@ -341,7 +413,7 @@ def network_hosts(self) -> list[str]:
     @property
     def port(self) -> int:
         """Return Port of OpenSearch unit."""
-        return 9200
+        return OPENSEARCH_HTTP_PORT
```
s-mlnk, line 416: I don't see any reason to keep it here instead of using this constant directly in OpensearchClient & Workload.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -651,3 +808,89 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
...
+        """Requests a new certificate with the given scope and type from the tls operator."""
+        self.charm.state.server.update({"tls_configured": None})
+        # TODO: Update peer cluster relation
+        # self.charm.tls.update_tls_flag_to_peer_cluster_relation("tls_configured", "remove")
+
+        for cert_type in [CertType.UNIT_HTTP, CertType.UNIT_TRANSPORT]:
+            csr = self.charm.state.secrets.get_object(Scope.UNIT, cert_type.val, peek=True)[
+                "csr"
+            ].encode("utf-8")
+            self.charm.tls_events.certs.request_certificate_revocation(csr)
```
s-mlnk, line 866: We can move `self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION, expiry_notification_time=23)` away from `TLSEventsHandler` to `OpenSearchBaseCharm` (or some another shared place) to decouple event handlers.

akram09: We did have a discussion with Rene regarding this and found that this is the middle way solution. Sine we cannot call functions between the event handlers and cannot move the requirer to OpenSearch events since it is scoped to tls events.

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+        with (
+            self.workload.temp_file(
+                mode="w+t", data=admin_secret["cert"], dir=self.workload.paths.conf
+            ) as tmp_cert,
+            self.workload.temp_file(
+                mode="w+t", data=admin_secret["key"], dir=self.workload.paths.conf
+            ) as tmp_key,
+        ):
+
+            self.opensearch_client.reload_tls_certificates(
```
reneradoi, line 462: We should have error handling here as the called method may raise.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+            self.charm.tls_manager.create_store_pwd_if_not_exists(
+                Scope.APP, CertType.APP_ADMIN, StoreType.KEYSTORE
+            )
+            csr = self.charm.tls_manager.create_certificate_signing_request(
+                Scope.APP, CertType.APP_ADMIN
+            )
+
+            if self.charm.state.tls_relation:
+                self.certs.request_certificate_creation(certificate_signing_request=csr)
+        elif not admin_cert.get("truststore-password"):
```
reneradoi: Having a condition here means we might silently fail (no certs requested despite a TLS relation created event). I believe we should either not safeguard it or have a proper measure in place (for instance a blocked status or deferral).

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+        unit_transport_csr = self.charm.tls_manager.create_certificate_signing_request(
+            Scope.UNIT, CertType.UNIT_TRANSPORT
+        )
+        unit_http_csr = self.charm.tls_manager.create_certificate_signing_request(
+            Scope.UNIT, CertType.UNIT_HTTP
+        )
+        if self.charm.state.tls_relation:
+            self.certs.request_certificate_creation(certificate_signing_request=unit_transport_csr)
+            self.certs.request_certificate_creation(certificate_signing_request=unit_http_csr)
+
```
reneradoi: Same comment: We should avoid silent failure.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+            return
+
+        try:
+            csr = self.charm.tls_manager.create_certificate_signing_request(
+                scope, cert_type, event.params.get("key", None), event.params.get("password", None)
+            )
+            if self.charm.model.get_relation(TLS_RELATION):
+                self.certs.request_certificate_creation(certificate_signing_request=csr)
+
```
reneradoi: Same here: We should avoid silent failure.

In this case we even have error handling in place already, so just need to add the exception.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+        scope = Scope.APP if cert_type == CertType.APP_ADMIN else Scope.UNIT
+        if scope == Scope.APP and not (
+            self.charm.unit.is_leader()
+            and self.charm.state.application.deployment_desc.typ
+            == DeploymentType.MAIN_ORCHESTRATOR
+        ):
+            event.log(
+                "Only the juju leader unit of the main orchestrator can set private key for the admin certificates."
+            )
+            return
```
reneradoi, line 95: Should we fail the event maybe? Otherwise it will be reported as successful to the user even though we didn't do anything.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+            certificate=event.certificate,
+            ca=event.ca,
+        )
+
+        current_stored_ca = self.charm.tls_manager.read_stored_ca()
+        if current_stored_ca != event.ca:
+            if not (deployment_desc := self.charm.state.application.deployment_desc):
+                logger.debug("Could not store new CA certificate.")
+                event.defer()
+                return
```
reneradoi: I'm curious: Why has this been added? For my understanding we should never get a certificate if the deployment description is not available yet because we would defer already on the tls-relation-created event (see line 98) and thereby not request a certificate.

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,354 @@
...
+            if not self.charm.tls_manager.store_new_ca(
+                self.charm.state.secrets.get_object(scope, cert_type.val, peek=True),
+                create_store_pwd=self.charm.unit.is_leader()
+                and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR,
+            ):
```
reneradoi, line 205: This condition is too complex, it's hard to read and understand. Why parametrize at all if this is the only place where it is used in the code (apart from unit tests)?

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+        if not (ca_trust_store.exists() and secrets):
+            return None
+
+        return self.read_ca(
+            alias=alias,
+            store_pwd=secrets.get("truststore-password"),
+            store_path=ca_trust_store,
+        )
+
+    def store_ca(
```
reneradoi: Why are we having a wrapper method around `_store_ca_chain`? Seems it's only passing through the parameters and not doing anything else - is this planned to be enhanced for anything? If not we can just remove it.

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+        """Common implementation to store a CA chain into a PKCS12 keystore."""
+        tmpdir = store_path.parent
+        starter_mode = "0664"
+        snap_user = "snap_daemon:root"
+        final_mode = "0640"
+        # import root first, then intermediates
+        certs = list(reversed(split_ca_chain(ca)))
+        if snap_user_with_write_permission and store_path.exists():
+            try:
+                self.workload.run_cmd(f"sudo chmod {starter_mode} {store_path}")
```
reneradoi: This will not go well in a container I guess. Another reason to move this to the workload's responsibility (as stated in my previous comment from the first review).

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+        try:
+            command = ""
+            if snap_user_with_write_permission:
+                command = (
+                    f"sudo chown {snap_user} {store_path}; sudo chmod {final_mode} {store_path};"
+                )
+            if add_read_perm:
+                command += f"sudo chmod +r {store_path}"
+            if command:
+                self.workload.run_cmd(command)
```
reneradoi: These commands will also fail in a K8s environment. I would not leave them here tbh.

opensearch_single_kernel/core/state.py
```diff
@@ -358,6 +430,83 @@ def get_unit(self, name: str):
...
+        """Handle internal flags during CA rotation routine."""
+        if not self.server.tls_ca_renewing:
+            # if the CA is not being renewed we don't have to do anything here
+            return
+
+        # if this flag is set, the CA rotation routine is complete for this unit
+        if self.server.tls_ca_renewed and self.ca_and_certs_rotation_complete_in_cluster():
+            # both CA rotation and certs rotation completed in the cluster
+            self.server.update({"tls_ca_renewing": None})
+            self.server.update({"tls_ca_renewed": None})
```
reneradoi: These should be unset with `update({"<key>": ""})` instead of `None`

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+        """Create a new chain.pem file for requests module"""
+        logger.debug("Updating requests TLS CA bundle")
+        admin_secret = self.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True)
+
+        # we store the pem format to make it easier for the python requests lib
+        chain_path = self.workload.paths.certs / "chain.pem"
+        parent_dir_path = chain_path.parent
+        if parent_dir_path:
+            parent_dir_path.mkdir(parents=True, exist_ok=True)
+        chain_path.write_text(admin_secret["chain"])
```
reneradoi: General comment: We need to have error handling for every instance of methods of the `PathProtocol`. They may raise a bunch of different errors.

reneradoi (commented):
Hey Akram, I tried to review more into detail this time. Please let me know if you have questions, I'm happy to have a session again to discuss the findings.

opensearch_single_kernel/managers/tls.py
akram09: Actually we are handling this exception at the TLSEventHandler level, or do you mean another scenario ?

opensearch_single_kernel/events/tls.py
akram09: Thank you I moved the check earlier in the function

opensearch_single_kernel/events/tls.py
akram09: Thank you

opensearch_single_kernel/events/tls.py
akram09: I totally agree with your point.

opensearch_single_kernel/events/tls.py
akram09: Thank you

opensearch_single_kernel/managers/tls.py
akram09: The thing is that in this PR it is not clear but once we have backups we would have another function that calls on store_ca with some different business logic

opensearch_single_kernel/managers/tls.py
akram09: I think this will be handled in the PR of k8s refactoring.

opensearch_single_kernel/managers/tls.py
akram09: Thank you

opensearch_single_kernel/managers/tls.py
reneradoi: Hm, not sure what my IDE tricked me into here but you are right, the error is already handled.

opensearch_single_kernel/managers/tls.py
reneradoi: Ack, then it makes sense.

opensearch_single_kernel/managers/tls.py
reneradoi: You can refer [here](https://github.com/canonical/valkey-operator/blob/9/edge/src/core/base_workload.py#L33) for an example of the error handling. This is the common implementation of `write_file` for both K8s and VM in the Valkey operator I have worked on these days. Similar but slightly different error handling should be done for other calls of `PathProtocol` (and also for pebble/subprocess execution btw).

akram09: I see your point. Will go for this approach. Thanks for sharing the example

coverage.xml
Mehdi-Bendriss, line 1: should this be committed?

opensearch_single_kernel/charms/base.py
```diff
@@ -47,6 +57,8 @@ def __init__(self, *args):
 
         # Event Handlers
         self.opensearch_events = OpenSearchEventsHandler(self)
+        self.tls_events = TLSEventsHandler(self)
+        self.actions_events = ActionsEventsHandler(self)
```
Mehdi-Bendriss: any strong reason not to have this moved to the tls events handler itself?

opensearch_single_kernel/core/models.py
```diff
@@ -125,7 +125,7 @@ class Node(Model):
     roles: list[str]
     ip: str
     app: App
-    unit_id: int
+    unit_number: int
```
Mehdi-Bendriss, line 128: i think unit_id is rather more common, no?

opensearch_single_kernel/core/state.py
```diff
@@ -358,6 +430,83 @@ def get_unit(self, name: str):
...
+                or "tls_ca_renewing" in self.peer_relation.data[unit]
+                or "tls_ca_renewed" in self.peer_relation.data[unit]
+            ):
+                return False
+        return True
+
+    @property
+    def ca_rotation_complete_in_cluster(self) -> bool:
+        """Check whether the CA rotation completed in all units."""
+        rotation_happening = False
```
Mehdi-Bendriss:
```suggestion
        rotation_in_progress = False
```

opensearch_single_kernel/core/state.py
```diff
@@ -358,6 +430,83 @@ def get_unit(self, name: str):
...
+        rotation_complete = True
+
+        # check peer units and current unit
+        if any([server.tls_ca_renewing for server in self.servers]):
+            rotation_happening = True
+        if not all([server.tls_ca_renewed for server in self.servers]):
+            rotation_complete = False
```
Mehdi-Bendriss: nit
```suggestion
        # check state of rotation on peer units and current unit
        rotation_in_progress = any([server.tls_ca_renewing for server in self.servers])
        rotation_complete = all([server.tls_ca_renewed for server in self.servers])
```

opensearch_single_kernel/events/actions.py
```diff
@@ -0,0 +1,112 @@
...
+            # TODO: Update relation of peer cluster provider
+            # self.peer_cluster_provider.refresh_relation_data(event)
+        except OpenSearchError as e:
+            event.fail(f"Failed changing the password: {e}")
+        except RuntimeError as e:
+            # From:
+            # https://github.com/canonical/operator/blob/ \
+            #     eb52cef1fba4df2f999f88902fb39555fb6de52f/ops/charm.py
+            if str(e) == "cannot defer action events":
+                event.fail("Cluster is not ready to update this password. Try again later.")
```
Mehdi-Bendriss: can you explain this? why do we consider deferrals here?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -72,9 +73,101 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+        # we need to wait for them all to be ready before deleting the old CA
+        if (
+            self.charm.tls_manager.read_stored_ca(self.charm.tls_manager.OLD_CA_ALIAS)
+            and self.charm.state.ca_and_certs_rotation_complete_in_cluster()
+        ):
+            logger.debug("update_status: Detected CA rotation complete in cluster")
+            self.charm.tls_manager.finalize_ca_certs_rotation()
+        # If relation not broken - leave
+        if self.charm.state.tls_relation is not None:
+            return
```
Mehdi-Bendriss: nit:
```suggestion
        if self.charm.state.tls_relation:
            return
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -72,9 +73,101 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+
+        # handle when/if certificates are expired
+        certs = self.charm.tls_manager.check_certs_expiration()
+        if certs:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/events/opensearch.py
```diff
@@ -72,9 +73,101 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            missing = [cert.val for cert in certs.keys()]
+            self.charm.status.set(
+                CharmStatuses.TLS_CERTS_EXPIRATION_ERROR,
+                dynamic_params={"certificates": ", ".join(missing)},
+            )
+
+            # stop opensearch in case the Node-transport certificate expires.
+            if certs.get(CertType.UNIT_TRANSPORT) is not None:
```
Mehdi-Bendriss: nit
```suggestion
            if certs.get(CertType.UNIT_TRANSPORT):
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -477,6 +573,70 @@ def _post_start_init(self, event: StartOpenSearch) -> None:
...
+
+        try:
+            self.stop_opensearch(restart=True)
+            logger.info("Restarting OpenSearch.")
+        except OpenSearchStopError as e:
+            logger.info(f"Error while Restarting Opensearch: {e}")
+            logger.exception(e)
+            self.charm.lock_manager.release()
+            event.defer()
+            self.charm.status.set(CharmStatuses.SERVICE_IS_STOPPING)
```
Mehdi-Bendriss: this doesn't seem needed

opensearch_single_kernel/events/opensearch.py
```diff
@@ -477,6 +573,70 @@ def _post_start_init(self, event: StartOpenSearch) -> None:
...
+            )
+            == 1
+        )
+        logger.debug("Restarting OpenSearch with ignore_lock=%s", ignore_lock)
+        self.charm.start_opensearch_event.emit(ignore_lock=ignore_lock)
+
+    def stop_opensearch(self, *, restart: bool = False) -> None:
+        """Stop OpenSearch service."""
+        self.charm.status.set(CharmStatuses.SERVICE_IS_STOPPING)
+        if self.charm.cluster_manager.opensearch_client.is_node_up():
```
Mehdi-Bendriss, line 630:
```suggestion

        if self.charm.cluster_manager.opensearch_client.is_node_up():
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -477,6 +573,70 @@ def _post_start_init(self, event: StartOpenSearch) -> None:
...
+                # TODO: Configure exclusions
+                if len(nodes) > 1:
+                    pass
+                # 1. Add current node to the voting + alloc exclusions
+                # self.opensearch_exclusions.add_current(voting=True, allocation=not restart)
+            except OpenSearchHttpError:
+                logger.debug("Failed to get online nodes, voting and alloc exclusions not added")
+
+        # block until all primary shards are moved away from the unit that is stopping
+        self.charm.health_manager.wait_for_shards_relocation()
```
Mehdi-Bendriss, line 645: for better readability
```suggestion
        self.charm.health_manager.wait_for_shards_relocation()
        
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -648,3 +807,89 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
             )
         else:
             return self.is_cluster_healthy_to_start()
+
+    def configure_tls_after_start(self):
```
Mehdi-Bendriss: (return type hints wherever they may be missing) 
```suggestion
    def _configure_tls_after_start(self) -> None:
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -648,3 +807,89 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
...
+        """Configure TLS state and certificates after OpenSearch is started."""
```
Mehdi-Bendriss: Very nice extraction from _post_start_init! 
Should we be more specific in the name that it concerns tls CA rotation?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -648,3 +807,89 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
...
+            key = secrets["key"].encode("utf-8")
+            key_password = secrets.get("key-password", None)
+            old_csr = secrets["csr"].encode("utf-8")
+            csr = self.charm.tls_manager.create_certificate_signing_request(
+                scope=Scope.UNIT,
+                cert_type=cert_type,
+                key=key,
+                password=key_password,
+                tls_file=False,
+            )
```
Mehdi-Bendriss, line 878: what do you think of having these defined in a method in the manager and simply getting a csr based on the cert_type?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -648,3 +807,89 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
...
+
+            self.charm.tls_events.certs.request_certificate_renewal(
+                old_certificate_signing_request=old_csr,
+                new_certificate_signing_request=csr,
+            )
+
+    def request_new_admin_certificate(self) -> None:
```
Mehdi-Bendriss, line 885: it does feel odd to see these methods in this events module

opensearch_single_kernel/events/opensearch.py
```diff
@@ -648,3 +807,89 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
...
+        """Request the generation of a new admin certificate."""
+        if not self.charm.unit.is_leader():
+            return
+        admin_secrets = (
+            self.charm.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True) or {}
+        )
+
+        key = admin_secrets["key"].encode("utf-8")
+        key_password = admin_secrets.get("key-password", None)
+        csr = self.charm.tls_manager.create_certificate_signing_request(
```
Mehdi-Bendriss, line 893: similar comment on the potential possibility to have these in a method in the manager

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,361 @@
...
+            event.fail(str(e))
+
+    def _on_tls_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Request certificate when TLS relation created."""
+        # TODO: Defer when upgrade is in progress
+        if not (deployment_desc := self.charm.state.application.deployment_desc):
+            event.defer()
+            return
+
+        if not self.charm.state.tls_relation:
```
Mehdi-Bendriss: how can this ever be if the elation has just been created?

opensearch_single_kernel/events/tls.py
```diff
@@ -0,0 +1,361 @@
...
+            scope, cert_type, secrets = self.charm.tls_manager.find_secret(
+                event.certificate_signing_request, "csr"
+            )
+            logger.debug(f"{scope.val}.{cert_type.val} TLS certificate available.")
+        except TypeError:
+            logger.debug("Unknown certificate available.")
+            return
+
+        # variables for better readability
+        is_unit_leader = self.charm.unit.is_leader()
```
Mehdi-Bendriss: nit
```suggestion
        is_leader_unit = self.charm.unit.is_leader()
```

opensearch_single_kernel/managers/cluster.py
```diff
@@ -215,16 +216,14 @@ def initialise_security_index(self):
...
-            # CertType.APP_ADMIN.val, peek=True)['truststore-password']}",
-            # "-tsalias ca",
-            # "-tst PKCS12",
-            # f"-ks {self.workload.paths.certs}/{CertType.APP_ADMIN}.p12",
-            # f"-kspass {self.state.secrets.get_object(Scope.APP,
-            #  CertType.APP_ADMIN.val, peek=True)['keystore-password']}",
-            # f"-ksalias {CertType.APP_ADMIN}",
-            # "-kst PKCS12",
+            f"-ts {self.workload.paths.certs}/ca.p12",
+            f"-tspass {self.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True)['truststore-password']}",
```
Mehdi-Bendriss:
```suggestion
            f"-tspass {admin_secrets['truststore-password']}",
```

opensearch_single_kernel/managers/cluster.py
```diff
@@ -215,16 +216,14 @@ def initialise_security_index(self):
...
+            "-tsalias ca",
+            "-tst PKCS12",
+            f"-ks {self.workload.paths.certs}/{CertType.APP_ADMIN}.p12",
+            f"-kspass {self.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True)['keystore-password']}",
```
Mehdi-Bendriss:
```suggestion
            f"-kspass {admin_secrets['keystore-password']}",
```

opensearch_single_kernel/managers/config.py
```diff
@@ -217,7 +220,59 @@ def add_seed_hosts(self, cm_ips: list[str]):
         # only update the file if there is data to update
         if cm_ips_set:
             lines = "\n".join([entry for entry in cm_ips_set if entry.strip()])
-            self.workload.paths.seed_hosts.write_text(f"{lines}\n")
+            self.workload.write_text(f"{lines}\n", self.workload.paths.seed_hosts)
+
+    def set_admin_tls_conf(self, secrets: dict[str, Any]):
```
Mehdi-Bendriss: nit: return type hints in this module

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+        if is_secret_found(u_http_secrets):
+            return Scope.UNIT, CertType.UNIT_HTTP, u_http_secrets
+
+        return None
+
+    def read_ca(self, alias: str, store_pwd: str, store_path: PathProtocol) -> str | None:
+        """Load stored CA cert."""
+        return (self.list_cas(store_pwd, store_path) or {}).get(alias)
+
+    def list_cas(
```
Mehdi-Bendriss: please consider than these methods will be used by:
- tls 
- s3 / backups 
- s3 / external static resources (later)

Probably worth adding a helpers_certificates or similar where functions have an arg: workload 
(applies to all other methods that are really helpers)

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+                args = f"{args} -passin pass:{key_pwd}"
+
+            try:
+                self.workload.run_cmd(cmd, args)
+                self.workload.run_cmd(f"sudo chmod +r {store_path}")
+            except OpenSearchCmdError as e:
+                logger.error("Error storing the TLS certificates for %s: %s", name, e)
+        logger.info("TLS certificate for %s stored.", name)
+
+    def update_request_ca_bundle(self) -> None:
```
Mehdi-Bendriss: can this be moved under the `add_ca_bundle` and maybe normalize both method names and maybe reuse internal logic?

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,773 @@
...
+            # TODO: Update peer cluster relation
+            # self.update_tls_flag_to_peer_cluster_relation("tls_configured", "add")
+
+    def get_cert_issuer(self, cert: str) -> str | None:
+        """Retrieve the certificate issuer from a string certificate."""
+        # to make sure the content is processed correctly by openssl, temporary store it in a file
+        with self.workload.temp_file(
+            mode="w+t", data=cert, dir=self.workload.root / "/tmp"
+        ) as tmp_ca_file:
+
```
Mehdi-Bendriss:
```suggestion
```

opensearch_single_kernel/workload/base.py
```diff
@@ -130,6 +139,61 @@ def paths(self) -> Paths:
...
+            path.write_text(content)
+        except (
+            FileNotFoundError,
+            LookupError,
+            NotADirectoryError,
+            PermissionError,
+            pathops.PebbleConnectionError,
+            ValueError,
+        ) as e:
+            raise OpenSearchFileOperationError(e)
```
Mehdi-Bendriss, line 159: is this caught somewhere in the caller code ?

opensearch_single_kernel/workload/vm.py
```diff
@@ -64,6 +68,36 @@ def install(self) -> None:
...
+    ):
+        """Create a temporary file and return the file, clean it once context is closed."""
+        f = tempfile.NamedTemporaryFile(
+            mode=mode, encoding=encoding, dir=dir, delete=False, errors=errors, suffix=suffix
+        )
+        file_path: PathProtocol = self.root / f.name
+        try:
+            if data:
+                self.write_text(data, file_path)
+            yield file_path
```
Mehdi-Bendriss, line 91: nice

opensearch_single_kernel/workload/vm.py
```diff
@@ -274,8 +308,26 @@ def run_cmd(
         except (TimeoutError, subprocess.TimeoutExpired):
             raise OpenSearchCmdError
 
+    @override
+    def stop(self) -> None:
+        """Stop the opensearch service."""
+        if not self.opensearch_snap.present:
```
Mehdi-Bendriss, line 314: realistically speaking, will this case ever happen from any caller? (since we should be crashing on_install if fails)

Mehdi-Bendriss (commented):
Thanks Akram! Really good work here! 
I have minor comments - the PR is already in a pretty good shape.
Can you confirm if the integration tests were ported 1:1 from the old repo?

opensearch_single_kernel/charms/base.py
akram09: Yes, I will move it to the tls events handler. Initially I thought that it would be better to group all actions in same event handler but it don't align with the approach we are following.

opensearch_single_kernel/core/models.py
akram09: In the start PR I changed the unit_number to unit_id. But Found that this impacted the `nodes_config` which can cause a problem if we want to keep it compatible with the `opensearch-operator` version

opensearch_single_kernel/events/actions.py
akram09: This part is mostly related to large deployments. There is a function that should be called here which updates the peer cluster relation and if we can not update the relations data we defer. I can omit this for the moment.

opensearch_single_kernel/events/opensearch.py
akram09: Why ?

opensearch_single_kernel/events/opensearch.py
akram09: I already have added the function `create_certificate_signing_request` that does exactly this. But the thing is that it is called from different perspectives. Sometimes we want to create a first csr sometimes we want to use secrets that are already existing so I wanted that function to be used in each case

opensearch_single_kernel/events/opensearch.py
akram09: Yes and this is mainly because we rely on the TLSCertificatesRequiresV3 that can only be initialized in event handlers and not at manager level. What we can do is to copy the functions that the tls_certificates have to update relation and have them in the cluster state, it would be a duplication but It would make things much much more clearer

opensearch_single_kernel/events/tls.py
akram09: Just a guard I will remove it.

opensearch_single_kernel/managers/tls.py
akram09: I am not a big fan of functions that takes args and are actually helpers methods. My idea would be that if we see that we have some shared code and really cannot use this manager I would add the helper to the BaseManager

opensearch_single_kernel/workload/vm.py
akram09: Yes it should not happen but still I would keep it as a guard if you don't mind

opensearch_single_kernel/workload/base.py
akram09: I added exception handling for this.

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: Because it is set first thing in `stop_opensearch` and to get into this `except` block `stop_opensearch` should have crashed at `self.charm.cluster_manager.stop_workload()` - meaning before the `SERVICE_STOPPED` status could have been set.

opensearch_single_kernel/managers/tls.py
Mehdi-Bendriss: I don't think bloating the BaseManager is the right approach, all unrelated managers will find themselves inheriting these methods unrelated to their specialization. 
Having a cleanly named helper module with the only dependency being the `workload` is not a code smell, and will not have any cyclic import being called exclusively from managers. 
Composition should allow for a more cleaner flow than inheritance in this case

opensearch_single_kernel/managers/tls.py
```diff
@@ -480,15 +488,34 @@ def _store_ca_chain(  # noqa: C901
...
+        if ca_chain is None:
+            admin_secret = self.state.secrets.get_object(
+                Scope.APP, CertType.APP_ADMIN.val, peek=True
+            )
+            ca_chain = admin_secret.get("chain")
+
+        # we store the pem format to make it easier for the python requests lib
+        chain_path = self.workload.paths.certs / "chain.pem"
+        parent_dir_path = chain_path.parent
+        if parent_dir_path:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/workload/vm.py
Mehdi-Bendriss: That's fine, but in general we should avoid being overly defensive in favor of having a predictable flow, with clarity on where things should or never fail (for readability and better reasoning)

Mehdi-Bendriss (commented):
Thanks Akram! Looks good! I only have a couple of unresolved comments I think still apply

opensearch_single_kernel/events/opensearch.py
akram09: I see your point, I thought you were mentioning the whole except block.

opensearch_single_kernel/managers/tls.py
akram09: Alright :+1:

opensearch_single_kernel/managers/tls.py
reneradoi: That is fine for me. We can add `# todo` comments at places where we know something still has to change, makes it easier for reviewer, too.

opensearch_single_kernel/workload/base.py
```diff
@@ -130,6 +139,61 @@ def paths(self) -> Paths:
...
+        try:
+            return path.read_text()
+        except (
+            FileNotFoundError,
+            LookupError,
+            NotADirectoryError,
+            PermissionError,
+            pathops.PebbleConnectionError,
+            ValueError,
+        ) as e:
```
reneradoi, line 177: For `pathops.read_text` the exceptions are different, see here: https://documentation.ubuntu.com/charmlibs/reference/charmlibs/pathops/#pathops.PathProtocol.read_text.

In this case, we can drop the Lookup-, Directory- and ValueErrors and need to add the Unicode one.

opensearch_single_kernel/workload/vm.py
```diff
@@ -64,6 +68,36 @@ def install(self) -> None:
...
+        suffix: str | None = None,
+    ):
+        """Create a temporary file and return the file, clean it once context is closed."""
+        f = tempfile.NamedTemporaryFile(
+            mode=mode, encoding=encoding, dir=dir, delete=False, errors=errors, suffix=suffix
+        )
+        file_path: PathProtocol = self.root / f.name
+        try:
+            if data:
+                self.write_text(data, file_path)
```
reneradoi, line 90: Error handling is missing here (or in the places where it's called), based on the `OpenSearchFileOperationError` thrown by `BaseWorkload.write_text`. E.g. if no permissions for the given path or encoding issues with certs, this would crash the operator here.

reneradoi (commented):
Just minor error handling comments from my side, rest looks good!

opensearch_single_kernel/workload/base.py
akram09: Fixed, Thank you Rene

opensearch_single_kernel/workload/vm.py
akram09: That is right, I have added exception handling at the events part since we can't do much in the managers most of the time

opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,781 @@
...
+        if ca_chain is None:
+            admin_secret = self.state.secrets.get_object(
+                Scope.APP, CertType.APP_ADMIN.val, peek=True
+            )
+            ca_chain = admin_secret.get("chain")
+
+        # we store the pem format to make it easier for the python requests lib
+        chain_path = self.workload.paths.certs / "chain.pem"
+        if parent_dir_path := chain_path.parent:
+            parent_dir_path.mkdir(parents=True, exist_ok=True)
```
gatici: It is better add exception handling here as this may raise if permissions are wrong.

opensearch_single_kernel/utils/config.py
```diff
@@ -266,7 +268,7 @@ def replace(
         if not path.exists():
             raise FileNotFoundError(f"{path} not found.")
 
-        data = path.read_text()
+        data = self.workload.read_text(path)
```
gatici, line 270: In line 266 we have a dead code as it is overriden in line 267.
 path = f"{self.base_path}{config_file}"
 path = self.base_path / config_file

gatici (dismissed):
Thank you Akram, Looks good!

opensearch_single_kernel/managers/tls.py
akram09: Thank you. Added mkdir to base workload and added exception handling.

opensearch_single_kernel/utils/config.py
akram09: Fixed

gatici (approved):

reneradoi (approved):
Awesome effort!

Mehdi-Bendriss (approved):
Thanks Akram! Great work!

PR #16, opened 2026-01-27
patch: OpenSearch Single Kernel Refactoring of Scale Up / Scale Down

akram09:
This PR refactors the following workflows to conform to the new charm architecture:

- Scale Up: The scale-up workflow has been refactored to handle the on_peer_relation_changed hook. It also incorporates logic previously located in the on_config_changed hook to support the configuration of the DeploymentDescription on an existing cluster.
- Scale Down: This workflow now consists of two main steps: on_peer_relation_departed and on_storage_detached.
- Exclusions: The OpenSearch Nodes Exclusion Manager has been refactored, and its functions are now integrated into the start and peer_relation hooks.

The following diagrams illustrate how these workflows now function.

image

image
image


opensearch_single_kernel/managers/tls.py
```diff
@@ -3,23 +3,782 @@
...
+        self,
+        scope: Scope,
+        cert_type: CertType,
+        secrets: dict[str, str] | None = None,
+        tls_file: bool = True,
+    ) -> bytes:
+        """Create CSR and save certificate key and password in secrets."""
+        if secrets:
+            key = secrets.get("key") if secrets.get("key") else None
+            password = secrets.get("key-password", None)
```
s-mlnk, line 210:
```suggestion
        key = secrets.get("key", None) if secrets else None
        password = secrets.get("key-password", None) if secrets else None
```
Will result in "var used before assignment" exception otherwise.

akram09: Yes, this is fixed now ! Thank you

opensearch_single_kernel/events/opensearch.py
```diff
@@ -819,6 +1120,25 @@ def unit_allowed_to_start(self, event: StartOpenSearch) -> bool:
         else:
             return self.is_cluster_healthy_to_start()
 
+    def trigger_peer_rel_changed(
```
reneradoi: We have the same method in `utils/helpers.py`

opensearch_single_kernel/events/opensearch.py
```diff
@@ -186,19 +417,31 @@ def _on_install(self, event: InstallEvent) -> None:
     def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """On config changed event. Useful for IP changes or for user provided config changes."""
         if self.charm.config_manager.update_host_if_needed():
-            # TODO: Handle TLS functions
-            pass
+            # This happens when the unit IP has changed
+            self.on_unit_ip_changed(event)
```
reneradoi, line 404: Previously there was logic for emitting a `RelationJoinedEvent` after IP address changes. Is this gone now for a specific reason?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -186,19 +417,31 @@ def _on_install(self, event: InstallEvent) -> None:
...
 
         if self.charm.unit.is_leader():
-            self.charm.cluster_manager.reconcile_cluster_config()
-            self.apply_status_from_deployment_desc(self.charm.state.application.deployment_desc)
+            if self.charm.cluster_manager.reconcile_cluster_config():
```
reneradoi: We could reduce one level of nesting here by combining this condition with `and` with the `leader` condition.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -555,15 +799,19 @@ def _post_start_init(self, event: StartOpenSearch) -> None:
             )
             self.charm.config_manager.cleanup_initial_cluster_managers()
 
-        self.charm.exclusions_manager.delete_current()
+        current_node = self.charm.config_manager.current_node
+        self.charm.exclusions_manager.delete_current(
+            node=current_node,
+            scope=Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
+        )
```
reneradoi, line 788:
```suggestion
        self.charm.exclusions_manager.delete_current(
            node=self.charm.config_manager.current_node,
            scope=Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
        )
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -633,11 +881,15 @@ def stop_opensearch(self, *, restart: bool = False) -> None:
                 # do not add exclusions if it's the last unit to stop
                 # otherwise cluster manager election will be blocked when starting up again
                 # and reusing storage
-                # TODO: Configure exclusions
                 if len(nodes) > 1:
-                    pass
-                # 1. Add current node to the voting + alloc exclusions
-                # self.opensearch_exclusions.add_current(voting=True, allocation=not restart)
+                    # 1. Add current node to the voting + alloc exclusions
+                    current_node = self.charm.config_manager.current_node
```
reneradoi: Same comment here as the earlier suggestion: `current_node` can be passed directly.

opensearch_single_kernel/managers/cluster.py
```diff
@@ -487,3 +652,19 @@ def stop_workload(self) -> None:
             time.sleep(3)
 
         self.state.server.update({"started": None})
```
reneradoi: Updates to the relation data should pass an empty string instead of `None` for deleting a field.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -790,7 +1042,56 @@ def _on_secret_changed(self, event: SecretChangedEvent) -> None:  # noqa: C901
...
+        logger.debug("Secret change for %s", str(label_key))
+
+        if is_leader and label_key == self.charm.state.secrets.password_key(KIBANA_SERVER_USER):
+            pass
+            # self.charm.opensearch_provider.update_dashboards_password()
+
+        # Non-leader units need to maintain local users in internal_users.yml
+        elif not is_leader and label_key in system_user_hash_keys:
+            password = event.secret.get_content()[label_key]
+            if sys_user := self.charm.state.secrets._user_from_hash_key(label_key):
```
reneradoi: I would rename the method `_user_from_hash_key` if it is accessed from outside.

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+                [
+                    node["node_name"]
+                    for node in resp["metadata"]["cluster_coordination"][
+                        "voting_config_exclusions"
+                    ]
+                ]
+            )
+        )
+
+    def remove_exclusions(self, exclusions: set[str], alt_hosts: list[str] | None = None) -> bool:
```
reneradoi: Suggestion: rename to `remove_voting_exclusions` for consistent naming.

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -4,21 +4,220 @@
...
+            logger.warning(f"Failed to fetch voting exclusions: {e}")
+            # no voting exclusion set
+            return {}
+
+    def _delete_voting(self, exclusions: set[str], scope: Scope) -> bool:
+        """Remove all voting exclusions and then re-adds the subset that should stay.
+
+        The API does not allow to remove a subset of the voting exclusions, at once.
+        """
+        current_excl = self._fetch_voting()
```
reneradoi: Why not using `opensearch_client.fetch_voting_config_exclusions()` directly?

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+        """Remove voting exclusions from OpenSearch cluster."""
```
reneradoi, line 105: For this and the other new methods in this class: Could you please add to the docstrings what the return value of each method means?

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+            resp = self.request("GET", "/_cluster/settings", alt_hosts=alt_hosts)
+            exclusions = resp["persistent"]["cluster"]["routing"]["allocation"]["exclude"]["_name"]
+            if exclusions:
+                allocation_exclusions = set(exclusions.split(","))
+        except KeyError:
+            # no allocation exclusion set
+            pass
+        finally:
+            return allocation_exclusions
+
```
reneradoi, line 158:
```suggestion
        try:
            resp = self.request("GET", "/_cluster/settings", alt_hosts=alt_hosts)
            exclusions = resp["persistent"]["cluster"]["routing"]["allocation"]["exclude"]["_name"]
            if exclusions:
                return set(exclusions.split(","))
        except KeyError:
                return set()
```

opensearch_single_kernel/common/constants.py
```diff
@@ -138,6 +138,9 @@ class OpenSearchPaths(BaseStrEnum):
 # OpenSearch default port
 OPENSEARCH_HTTP_PORT = 9200
 
+# OpenSearch storage name
+OPENSEARCH_STORAGE_NAME = "opensearch-data"
```
reneradoi: Just a heads-up: While we started with just the data storage, there have been multiple different storages on some products recently (e.g. also for backups, logs). Would be a good chance to name this literal something with `DATA` right away to not have to rename it sometime later.

opensearch_single_kernel/managers/cluster.py
```diff
@@ -445,6 +581,35 @@ def recompute_nodes_conf(self, app_id: str, nodes: list[Node]) -> dict[str, Node
         )
         return nodes_by_name
 
+    def cleanup_on_last_unit_removal(self):
+        """Clean up cluster state on last unit removal."""
+        if self.state.peer_relation:
+            self.state.application.update({"bootstrap_contributors_count": None})
```
reneradoi: Applies here too: `update` expects `dict[str, str]`.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+                ]
+                self.charm.cluster_manager.compute_and_broadcast_updated_topology(remaining_nodes)
+            elif self.charm.app.planned_units() == 0:
+                # This is the last unit being removed
+                # We want to clean things up in case of a cold start later
+                self.charm.cluster_manager.cleanup_on_last_unit_removal()
+
+            # No cluster managers left in the cluster fleet
+            # raise so we do not lose the cluster state
+            # TODO:
```
reneradoi: What's the todo here?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+                # check cluster status
+                if self.charm.cluster_manager.alt_hosts:
+                    health_color = self.charm.status.apply_health(
+                        wait_for_green_first=True, use_localhost=False, unit=False
+                    )
+                    if health_color == HealthColors.RED:
+                        raise OpenSearchHAError(CharmStatuses.CLUSTER_HEALTH_RED.value.message)
+                else:
+                    raise OpenSearchHAError(CharmStatuses.CLUSTER_HEALTH_UNKNOWN.value.message)
+        finally:
```
reneradoi: We could invert here for better readability: 
```
if not alt_hosts:
    raise

the rest here
```

opensearch_single_kernel/managers/config.py
```diff
@@ -306,3 +317,79 @@ def set_jvm_heap_size(self, heap_size_in_kb: int) -> None:
...
+            conf = self.yaml_setter.load(self.CONFIG_YML)
+
+            # also, if possible we rely on the Deployment Description (databag)
+            deployment_desc = self.state.application.deployment_desc
+
+            # Application Priority: Deployment Description
+            # Reason: No reason to re-construct the App object
+            #  - it's available 99% of scenarios
+            #  - it's the same object as a re-constructed one (i.e. no dynamic changes on App)
+            if deployment_desc is None:
```
reneradoi:
```suggestion
if not (deployment_desc: = self.state.application.deployment_desc):
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            elif self.charm.app.planned_units() == 0:
+                # This is the last unit being removed
+                # We want to clean things up in case of a cold start later
+                self.charm.cluster_manager.cleanup_on_last_unit_removal()
+
+            # No cluster managers left in the cluster fleet
+            # raise so we do not lose the cluster state
+            # TODO:
+
+            # we attempt to flush the translog to disk
```
reneradoi:
```suggestion
        # we attempt to flush the translog to disk
```

reneradoi (commented):
Nice work Akram! Not many design comments this time, looks really good. I've made some suggestions here and there for clarity and readability, and left one questions on the logic flow. Other than that it's not far from being ready.

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
         self.workload = workload
         self.admin_secret = admin_secret
 
+    def apply_no_replication_to_index(
```
Mehdi-Bendriss: this is not an accurate description of this call, `0-all` means: auto replica scaling, meaning if the cluster had 1 node, the replica count will be 0, if there were 5 nodes however - the replica count would be 4.

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+        index: str,
+    ) -> None:
+        """Apply replication settings to an index."""
+        self.request(
+            method="PUT",
+            endpoint=f"/{index}/_settings",
+            payload={"index": {"auto_expand_replicas": "0-all"}},
+        )
+
+    def fetch_voting_config_exclusions(self, alt_hosts: list[str] | None = None) -> set[str]:
```
Mehdi-Bendriss:
```suggestion
    def fetch_voting_exclusions(self, alt_hosts: list[str] | None = None) -> set[str]:
```

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+            retries=3,
+        )
+        if response >= 400:
+            logger.debug("Failed to add voting exclusions, response %s", response)
+            return False
+
+        logger.debug("Added voting exclusions for:  %s", exclusions)
+        return True
+
+    def fetch_allocations(self, alt_hosts: list[str] | None = None) -> set[str]:
```
Mehdi-Bendriss:
```suggestion
    def fetch_allocation_exclusions(self, alt_hosts: list[str] | None = None) -> set[str]:
```

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+        """Apply replication settings to an index."""
+        self.request(
+            method="PUT",
+            endpoint=f"/{index}/_settings",
+            payload={"index": {"auto_expand_replicas": "0-all"}},
+        )
+
+    def fetch_voting_config_exclusions(self, alt_hosts: list[str] | None = None) -> set[str]:
+        """Fetch the voting exclusions config."""
+        resp = self.request(
```
Mehdi-Bendriss: I thought a bit about the current module, and I feel like the exception handling in this module was designed to "fit" the ported caller code. Sometimes we let the exception propagate, sometimes we catch it and return an opinionated value.

We should revisit this, take a look at the client code and try to have a uniform way of accessing this client class and handling exceptions and be opinionated about the return values. It won't be always possible, but we should do that as much as possible.

(as an example, we let the KeyError a few lines below propagate, while we catch it in `fetch_allocations`)

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
         self.workload = workload
         self.admin_secret = admin_secret
 
+    def apply_no_replication_to_index(
+        self,
+        index: str,
+    ) -> None:
+        """Apply replication settings to an index."""
+        self.request(
```
Mehdi-Bendriss, line 72: where do we handle the OpenSearchHttpError?

opensearch_single_kernel/common/client.py
Mehdi-Bendriss: and walrus on the `exclusions`

opensearch_single_kernel/common/client.py
```diff
@@ -58,6 +156,20 @@ def get_node_id(self, unit_name: str) -> str | None:
             if node["name"] == unit_name:
                 return n_id
```
Mehdi-Bendriss, line 196: missing return None

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,103 @@ def __init__(
...
+    ) -> bool:
+        """Register new allocation exclusions."""
+        try:
+            existing = set() if override else self.fetch_allocations(alt_hosts=alt_hosts)
+            all_allocs = existing.union(allocations if allocations is not None else {node.name})
+            response = self.request(
+                "PUT",
+                "/_cluster/settings",
+                {"persistent": {"cluster.routing.allocation.exclude._name": ",".join(all_allocs)}},
+                alt_hosts=alt_hosts,
```
Mehdi-Bendriss: please review the retry mechanism in all the methods of this class - related to a a comment in a previous PR, we should be uniform in the retry mechanism.

All methods should have a retry, some currently don't like `get_current_node`, `get_roles` etc

opensearch_single_kernel/common/constants.py
```diff
@@ -181,3 +184,12 @@ class OpenSearchPaths(BaseStrEnum):
 # Messages
 PEER_CLUSTER_NO_RELATION = "Cannot start. Waiting for peer cluster relation..."
 PEER_CLUSTER_WRONG_RELATION = "Cluster name doesn't match with related cluster. Remove relation."
+CLUSTER_MANAGER_VOTING_ROLES_PROVIDED_INVALID = (
+    "cluster_manager and voting_only roles cannot be both set on the same nodes."
```
Mehdi-Bendriss:
```suggestion
    "cluster_manager and voting_only roles cannot be both set on the same node."
```

opensearch_single_kernel/common/constants.py
```diff
@@ -181,3 +184,12 @@ class OpenSearchPaths(BaseStrEnum):
 # Messages
 PEER_CLUSTER_NO_RELATION = "Cannot start. Waiting for peer cluster relation..."
 PEER_CLUSTER_WRONG_RELATION = "Cluster name doesn't match with related cluster. Remove relation."
+CLUSTER_MANAGER_VOTING_ROLES_PROVIDED_INVALID = (
```
Mehdi-Bendriss, line 188: In general, it is okay to use the `CM` abreviation for cluster_manager

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            self._on_opensearch_data_storage_detaching,
+        )
 
         # --- OpenSearch Custom events ---
         self.framework.observe(self.charm.start_opensearch_event, self._on_start_opensearch)
         self.framework.observe(self.charm.restart_opensearch_event, self._on_restart_opensearch)
 
+        # Ensure that only one instance of the `_on_peer_relation_changed` handler exists
+        # in the deferred event queue
+        self._is_peer_rel_changed_deferred = False
```
Mehdi-Bendriss: please see if this is really useful

opensearch_single_kernel/managers/cluster.py
```diff
@@ -85,18 +88,45 @@ def _is_connected():
...
+
+        # TODO: Should we add an entry on DeploymentDesc "errors" to reflect on status?
+        self.state.application.put_object("deployment-description", deployment_desc.to_dict())
+
+        # TODO: once peer clusters relation implemented, we should apply all directives
+        #  + removing them from queue. We currently only apply the status.
+        return True
 
-        # TODO: Handle existing cluster case
+    def recompute_roles_if_needed(self):
```
Mehdi-Bendriss: we no longer recompute node roles, why do we still need this?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+
+            if health in [HealthColors.UNKNOWN, HealthColors.YELLOW_TEMP]:
+                # we defer because we want the temporary status to be updated
+                logger.debug("Cluster health temp yellow or unknown. Deferring event.")
+                event.defer()
+                # If the handler is called again within this Juju hook, we will abandon the event
+                self._is_peer_rel_changed_deferred = True
+
+        # we want to have the most up-to-date info broadcasted to related sub-clusters
+        # if self.opensearch_peer_cm.is_provider():
```
Mehdi-Bendriss, line 161: Do we expect this to still be in the opensearch events handler when we implement large deployments?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+        # self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
+
+        # update any orchestrators about planned units
+        # if self.opensearch_peer_cm.is_consumer():
+        # self.peer_cluster_requirer.refresh_requirer_relation_data()
+
+        # for relation in self.model.relations.get(ClientRelationName, []):
+        # self.opensearch_provider.update_endpoints(relation)
+
+        # register new cm addresses on every node
```
Mehdi-Bendriss, line 171: Could we clarify why we do this here? i believe this is in response to the `started` flag being set on the unit peer rel after a unit finishes its start

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            # to relate integrators to this new main orchestrator
+            # self.peer_cluster_provider.check_credentials_with_missing_relations()
+            # if self.model.relations[PeerClusterRelationName]:
+            # self.peer_cluster_requirer.apply_orchestrator_status()
+        elif event.relation.data.get(event.app):
+            # if app_data + app_data["nodes_config"]: Reconfigure + restart node on the unit
+            if self.charm.config_manager.reconfigure_unit():
+                self.charm.status.set(CharmStatuses.WAITING_TO_START)
+                logger.debug("Restarting opensearch due to reconfiguring node roles")
+                self.charm.restart_opensearch_event.emit()
```
Mehdi-Bendriss, line 189: can we please rethink this flow if still applies?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+
+        # check requirements
+        if self.charm.state.application.deployment_desc:
```
Mehdi-Bendriss: I don't understand this check, should we have this at the beginning of the hook? 
if yes, should we discard this hook completely if not yet available?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            self.check_profile_missing_requirements()
+
+        if not (unit_data := event.relation.data.get(event.unit)):
+            return
+
+        if self.charm.state.application.deployment_desc:
```
Mehdi-Bendriss: related to the above comment

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: Should this be at the beginning of the hook? If yes, should we discard the event if None?

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: related to a previous comment

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            current_node = self.charm.config_manager.current_node
+            self.charm.exclusions_manager.cleanup(
+                Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
+                current_node,
+            )
```
Mehdi-Bendriss:
```suggestion
            self.charm.exclusions_manager.cleanup(
                Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
                self.charm.config_manager.current_node,
            )
```

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+
+        if self.charm.unit.is_leader() and unit_data.get("bootstrap_contributor"):
+            contributor_count = self.charm.state.application.bootstrap_contributors_count
+            self.charm.state.application.bootstrap_contributors_count = contributor_count + 1
```
Mehdi-Bendriss, line 203: For a future PR (please create an issue), let's rethink the actual bootstrap_contributors logic

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
 
         # --- OpenSearch Custom events ---
         self.framework.observe(self.charm.start_opensearch_event, self._on_start_opensearch)
         self.framework.observe(self.charm.restart_opensearch_event, self._on_restart_opensearch)
 
+        # Ensure that only one instance of the `_on_peer_relation_changed` handler exists
+        # in the deferred event queue
+        self._is_peer_rel_changed_deferred = False
+
+    def _on_peer_relation_created(self, event: RelationCreatedEvent):
```
Mehdi-Bendriss: missing return type hint in this hook and subsequent ones

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            node
+            for node in self.charm.cluster_manager.get_nodes(True)
+            if node.name != format_unit_name(event.departing_unit, app=current_app)
+        ]
+
+        self.charm.status.apply_health(wait_for_green_first=True, unit=False)
+
+        n_units = sum(1 for node in remaining_nodes if node.app.id == current_app.id)
+        if n_units == self.charm.app.planned_units():
+            self.charm.cluster_manager.compute_and_broadcast_updated_topology(remaining_nodes)
```
Mehdi-Bendriss, line 240: We need to start seriously thinking about this method.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+
+            # No cluster managers left in the cluster fleet
+            # raise so we do not lose the cluster state
+            # TODO:
+
+            # we attempt to flush the translog to disk
+        self.charm.cluster_manager.flush_translog_to_disk()
+
+        try:
+            self.stop_opensearch()
```
Mehdi-Bendriss: should this be in the events handler?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,11 +90,231 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            logger.debug(
+                f"Waiting for units to leave: expecting {self.charm.app.planned_units()}, currently {n_units}. Deferring event."
+            )
+            event.defer()
+        self.charm.exclusions_manager.add_to_cleanup_list(
+            unit_name=format_unit_name(event.departing_unit.name, deployment_desc.app),
+            scope=Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
+        )
+
+    def _on_opensearch_data_storage_detaching(self, event: StorageDetachingEvent):  # noqa: C901
```
Mehdi-Bendriss: I feel like this method wasn't exactly refactored 😆 unless not needed

opensearch_single_kernel/events/opensearch.py
```diff
@@ -555,15 +799,19 @@ def _post_start_init(self, event: StartOpenSearch) -> None:
             )
             self.charm.config_manager.cleanup_initial_cluster_managers()
 
-        self.charm.exclusions_manager.delete_current()
+        current_node = self.charm.config_manager.current_node
+        self.charm.exclusions_manager.delete_current(
+            node=current_node,
```
Mehdi-Bendriss:
```suggestion
        self.charm.exclusions_manager.delete_current(
            node=self.charm.config_manager.current_node,
```

opensearch_single_kernel/managers/cluster.py
```diff
@@ -445,6 +581,35 @@ def recompute_nodes_conf(self, app_id: str, nodes: list[Node]) -> dict[str, Node
...
+        #    self.peers_data.delete(Scope.APP, "orchestrators")
+        #    self.peer_cluster_provider.clean_all_relation_data()
+        # elif self.opensearch_peer_cm.is_consumer():
+        #    self.peer_cluster_requirer.refresh_requirer_relation_data()
+
+    def flush_translog_to_disk(self) -> None:
+        """Flush OpenSearch translog to disk."""
+        if self.opensearch_client.is_node_up():
+            try:
+                self.opensearch_client.request("POST", "/_flush")
```
Mehdi-Bendriss: well why have the client class if we're still exposing endpoints? and we also need a retry here

opensearch_single_kernel/managers/cluster.py
```diff
@@ -487,3 +652,19 @@ def stop_workload(self) -> None:
             time.sleep(3)
 
         self.state.server.update({"started": None})
+
+    def apply_upstream_fixes(self):
```
Mehdi-Bendriss:
```suggestion
    def apply_upstream_fixes(self) -> None:
```

opensearch_single_kernel/managers/config.py
```diff
@@ -306,3 +317,79 @@ def set_jvm_heap_size(self, heap_size_in_kb: int) -> None:
             f"-Xmx{str(heap_size_in_kb)}k",
             regex=True,
         )
+
+    def add_cm_addresses_to_conf(self):
```
Mehdi-Bendriss: missing return type hints (please inspect the modules)

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -4,21 +4,220 @@
...
+                    raise OpenSearchExclusionsException("Failed to delete voting exclusion.")
+
+        if allocation and node.is_data():
+            if not self._delete_allocations(node):
+                logger.error(f"Failed to delete shard allocation exclusion: {node.name}.")
+                # Load the content of the list, avoiding '' entries
+                if scope == Scope.APP:
+                    state = self.state.application
+                else:
+                    state = self.state.server
```
Mehdi-Bendriss: ternary operator

opensearch_single_kernel/managers/lock.py
```diff
@@ -283,7 +283,7 @@ def acquired(self) -> bool:  # noqa: C901
         if host or alt_hosts:
             logger.debug("[Node lock] 1+ opensearch nodes online")
             try:
-                online_nodes = len(self.nodes(use_localhost=host is not None, hosts=alt_hosts))
+                online_nodes = len(self._nodes(use_localhost=host is not None, hosts=alt_hosts))
```
Mehdi-Bendriss, line 287: This module isn't refactored yet, correct?

opensearch_single_kernel/managers/tls.py
```diff
@@ -358,6 +358,12 @@ def store_new_tls_resources(self, cert_type: CertType, secrets: dict[str, Any])
             key_pwd=secrets.get("key-password"),
         )
 
+    def delete_stored_tls_resources(self):
```
Mehdi-Bendriss:
```suggestion
    def delete_stored_tls_resources(self) -> None:
```

opensearch_single_kernel/workload/vm.py
```diff
@@ -122,6 +122,14 @@ def get_host_public_ip(self) -> str | None:
 
         return output.out.strip()
 
+    def is_started(self) -> bool:
+        """Check if OpenSearch is started."""
+        reachable = self.is_reachable(self.host, self.port)
+        if not reachable:
```
Mehdi-Bendriss: walrus

last_written_value
Mehdi-Bendriss, line 1: needs to be added to the .gitignore

Mehdi-Bendriss (commented):
Thanks Akram! Good work! 
Let's please watch the content of the `opensearch` event handler to not reach the size we are at in the `opensearch_base_charm` of the other repo :)

Some flows need to be looked at to see if there is some refactoring to be done there, or if we were already good.

opensearch_single_kernel/events/opensearch.py
akram09: Thank you for the catch !

opensearch_single_kernel/managers/exclusions.py
akram09: I personally find it more readable and we can reuse that code in case of need.

opensearch_single_kernel/common/client.py
akram09: Yes, I will. Thank you

opensearch_single_kernel/common/constants.py
akram09: Thank you !

opensearch_single_kernel/events/opensearch.py
akram09: This part is mainly large deployment handling.

opensearch_single_kernel/common/client.py
akram09: I just checked the documentation it makes more sense.

opensearch_single_kernel/common/client.py
akram09: Why ?

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: Update on this, this is no longer needed, ops now de-duplicates deferred events. We just need to make sure we have bumped the version of ops to the newest one

opensearch_single_kernel/common/client.py
akram09: You are right, I missed that. My approach would be that we only propagate charm exceptions from this module. Meaning that if we can not handle an exception at this level we throw a custom exception that the manager/event handler can handle. This will provide a good abstraction between the client and the callers methods.

opensearch_single_kernel/common/client.py
akram09: Should we block the start process if this don't occures ?

opensearch_single_kernel/common/client.py
akram09: I agree that all methods should have a retry but having the same retry values would make the charm very slow I think.

opensearch_single_kernel/events/opensearch.py
akram09: I am planning to move most of these to the state and calls from manager.

opensearch_single_kernel/events/opensearch.py
akram09: I think this is the main (only) place we are actually setting the seed hosts after we have peer units in the cluster.

opensearch_single_kernel/events/opensearch.py
akram09: I believe this is a bit tight to large deployment. Would create an issue for this to be handled by the large deployment PR. What do you think ?

opensearch_single_kernel/events/opensearch.py
akram09: In theory it should not have a big effect since if we have peer relation changed event this means we have already went through the config changed and configured deployment_desc since this is the main attribute that is set at first. 
I will test this quickly and let you know

opensearch_single_kernel/common/client.py
gatici: I can suggest to rename like ` fetch_voting_exclusions_config`.

opensearch_single_kernel/common/client.py
```diff
@@ -43,6 +44,114 @@ def __init__(
...
+                        ]
+                    ]
+                )
+            )
+        except KeyError:
+            # no voting exclusions set
+            return set()
+
+    def remove_voting_exclusions(
+        self, exclusions: set[str], alt_hosts: list[str] | None = None
```
gatici: exclusions parameters are not used here other than logging purpose.

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -4,21 +4,265 @@
...
+
+        Returns:
+            A set of unit names that are currently in the voting exclusion list.
+        """
+        try:
+            return self.opensearch_client.fetch_voting_config_exclusions(alt_hosts=self.alt_hosts)
+        except OpenSearchHttpError as e:
+            logger.warning(f"Failed to fetch voting exclusions: {e}")
+            # no voting exclusion set
+            return {}
```
gatici: Return type expects set() not a dict.

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -4,21 +4,265 @@
...
+            to_stay = current_excl - exclusions
+        if current_excl == to_stay:
+            # Nothing to do
+            logger.debug("No voting exclusions to delete, current set is %s", to_stay)
+            return True
+
+        # "wait_for_removal" is VERY important, it removes all voting configs immediately
+        # and allows any node to return to the voting config in the future
+        try:
+            if not self.opensearch_client.remove_exclusions(
```
gatici: remove_exclusions is not defined.

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -4,21 +4,265 @@
...
+        """Add Voting and alloc exclusions for a target unit.
+
+        This method is just a clean-up-later routine. We (re)add the unit to the exclusions and,
+        hence, the leader of this app will be aware of this unit's removal and log it into its
+        app-level peer data.
+        """
+        state = self.state.application if scope == Scope.APP else self.state.server
+        for lst in [state.allocation_exclusions_to_delete, state.delete_voting_exclusions]:
+            # Load the content of the list, avoiding '' entries
+            lst = lst.union({unit_name})
```
gatici: this adds unit_name to a local variable but it does not assign it back to state object.

opensearch_single_kernel/managers/lock.py
akram09: Yes, it is still not refactored

opensearch_single_kernel/common/constants.py
akram09: I still prefer clarity, but I can update if you want

opensearch_single_kernel/events/opensearch.py
akram09: Created an issue for it . [Ref](https://github.com/canonical/opensearch-single-kernel-library/issues/26)

opensearch_single_kernel/events/opensearch.py
akram09: I want to do that once implementing large deployment since I am still missing some knowledge on that side

opensearch_single_kernel/events/opensearch.py
akram09: That function is interacting with multiple managers. We can move it to the charm. Wdu think ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,12 +95,236 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+                # We already deferred this event during this Juju event. Retry on the next
+                # Juju event.
+                return
+
+            if health in [HealthColors.UNKNOWN, HealthColors.YELLOW_TEMP]:
+                # we defer because we want the temporary status to be updated
+                logger.debug("Cluster health temp yellow or unknown. Deferring event.")
+                event.defer()
+                # If the handler is called again within this Juju hook, we will abandon the event
+                self._is_peer_rel_changed_deferred = True
```
gatici: I am afraid, this may stay forever as  setting this to False is only done in _init_ method which run only once.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -79,12 +95,236 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+
+        if not (unit_data := event.relation.data.get(event.unit)):
+            return
+
+        self.charm.exclusions_manager.cleanup(
+            Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
+            node=self.charm.config_manager.current_node,
+        )
+
+        if self.charm.unit.is_leader() and unit_data.get("bootstrap_contributor"):
```
gatici, line 201: as this event is deferred, deferred calls may increment the bootstrap_contributors_count.

opensearch_single_kernel/managers/cluster.py
```diff
@@ -162,7 +179,113 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
...
-    def _user_config(self):
+    def _existing_cluster_setup(
+        self, config: PeerClusterConfig, prev_deployment: DeploymentDescription
+    ) -> DeploymentDescription:
+        """Build deployment description of an existing (started or not) cluster."""
+        logger.debug(
+            "Found deployment description using existing cluster setup. deployment desc: %s",
+            prev_deployment,
+        )
+        directives = prev_deployment.pending_directives
```
gatici: directive just assign a reference to prev_deployment and all appends and removals will modify prev_deployment. We should not touch the prev_deployment and create a copy of it.

opensearch_single_kernel/managers/cluster.py
```diff
@@ -162,7 +179,113 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
...
+            deployment_state = DeploymentState(value=State.ACTIVE, message="")
+            directives.append(Directive.SHOW_STATUS)
+            directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
+
+        dep_type = deployment_type(config, start_mode, prev_deployment.typ)
+        return DeploymentDescription(
+            app=prev_deployment.app,
+            config=PeerClusterConfig(
+                cluster_name=prev_deployment.config.cluster_name,
+                init_hold=prev_deployment.config.init_hold,
```
gatici: this is always set to prev_deployment.init_hold. config.init_hold never considered. Is this desired behaviour ?

opensearch_single_kernel/managers/cluster.py
```diff
@@ -162,7 +179,113 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
...
+                roles=config.roles,
+                data_temperature=config.data_temperature,
+            ),
+            start=start_mode,
+            state=deployment_state,
+            typ=dep_type,
+            pending_directives=list(set(directives)),
```
gatici, line 232: This does not produce deterministic result as set does not protect order of objects.

gatici (commented):
@akram09 Thank you Akram! This is huge work. I added some comments to clarify intended behavior.

opensearch_single_kernel/events/opensearch.py
akram09: Fixed

opensearch_single_kernel/events/opensearch.py
akram09: This is the last point of the function. We are not defering after this

opensearch_single_kernel/managers/cluster.py
akram09: I am assuming yes, I think this part will be addressed by another PR regarding external config

opensearch_single_kernel/managers/cluster.py
akram09: I think its okey as long as we are saving the deployment desc in the right order

opensearch_single_kernel/common/client.py
akram09: This is catched in `apply_upstream_fixes()`

opensearch_single_kernel/common/client.py
akram09: Fixed now :+1:

opensearch_single_kernel/events/opensearch.py
akram09: Thank you

opensearch_single_kernel/managers/cluster.py
akram09: Updated

opensearch_single_kernel/events/opensearch.py
akram09: Moved to the charm

opensearch_single_kernel/events/opensearch.py
akram09: Will go through it again :+1:

opensearch_single_kernel/common/client.py
```diff
@@ -98,6 +128,7 @@ def add_voting_exclusions(
             alt_hosts=alt_hosts,
             resp_status_code=True,
             retries=3,
+            wait_strategy=wait_exponential(min=2),
```
Mehdi-Bendriss, line 131: I don't think we need an exponential wait, we can get away with a fixed wait for all operations, or where do you see value in this being passed as an argument on top of the `retries` one?

opensearch_single_kernel/common/client.py
```diff
@@ -44,48 +44,78 @@ def __init__(
...
             resp_status_code=True,
+            retries=3,
+            wait_strategy=wait_exponential(min=2),
         )
         if response >= 400:
             logger.debug("Failed to remove voting exclusions, response %s", response)
             return False
 
-        logger.debug("Removed voting for:  %s", exclusions)
+        logger.debug("Removed voting exclusions.")
```
Mehdi-Bendriss, line 118: Could you create a new issue? i would like us to revisit all the log levels we have in the charm
And for new PRs, let's pay attention to the log levels we use for each line

opensearch_single_kernel/common/client.py
```diff
@@ -44,48 +44,78 @@ def __init__(
         self.workload = workload
         self.admin_secret = admin_secret
 
-    def apply_no_replication_to_index(
+    def flush_opensearch_translog(self, alt_hosts: list[str] | None = None) -> None:
```
Mehdi-Bendriss: i think we can drop the `opensearch` in the method name

opensearch_single_kernel/common/client.py
```diff
@@ -44,48 +44,78 @@ def __init__(
...
-                ]
+                )
             )
-        )
+        except KeyError:
+            # no voting exclusions set
+            return set()
 
-    def remove_exclusions(self, exclusions: set[str], alt_hosts: list[str] | None = None) -> bool:
+    def remove_exclusions(self, alt_hosts: list[str] | None = None) -> bool:
```
Mehdi-Bendriss: `remove_voting_exclusions`

opensearch_single_kernel/common/client.py
```diff
@@ -106,19 +137,24 @@ def add_voting_exclusions(
...
+                return set(exclusions.split(","))
         except KeyError:
-            # no allocation exclusion set
             pass
-        finally:
-            return allocation_exclusions
+
+        return set()
 
     def add_allocations(
```
Mehdi-Bendriss: `add_allocation_exclusions`

opensearch_single_kernel/common/client.py
```diff
@@ -128,18 +164,17 @@ def add_allocations(
...
-                "PUT",
-                "/_cluster/settings",
-                {"persistent": {"cluster.routing.allocation.exclude._name": ",".join(all_allocs)}},
-                alt_hosts=alt_hosts,
-            )
-            return "acknowledged" in response
-        except OpenSearchHttpError:
-            return False
+        existing = set() if override else self.fetch_allocation_exclusions(alt_hosts=alt_hosts)
+        all_allocs = existing.union(allocations if allocations is not None else {node.name})
```
Mehdi-Bendriss: `all_exclusions = existing.union(exclusions or {node.name})`

in general the concept of allocation_exclusion is rather non intuitive as it is a negative (i.e: exclusion), it is in fact the opposite of "allocation" 

(github won't allow me for some reason to make a code suggestion)

opensearch_single_kernel/common/constants.py
Mehdi-Bendriss: Yes please, as it is too long. For statuses we can potentially keep as is, but in the code (variables) i'd rather we get used to `cm`

opensearch_single_kernel/common/client.py
```diff
@@ -387,6 +437,7 @@ def request(  # noqa
         check_hosts_reach: bool = True,
         resp_status_code: bool = False,
         retries: int = 1,
+        wait_strategy: WaitBaseT = wait_fixed(1),
```
Mehdi-Bendriss, line 440: related to a previous question, what is the value in having different wait strategies on top of a configurable retry count?

opensearch_single_kernel/core/models.py
```diff
@@ -33,8 +32,6 @@ class Model(ABC, BaseModel):
     """Base model class."""
 
     def __init__(self, **data: Any) -> None:
-        if self.__custom_root_type__ and data.keys() != {ROOT_KEY}:
-            data = {ROOT_KEY: data}
```
Mehdi-Bendriss, line 37: I cannot remember why, but i believe this was needed as the deserialized pydantic models were coming with a `__root__` field

opensearch_single_kernel/core/models.py
```diff
@@ -99,16 +96,17 @@ class App(Model):
...
         """Generate the attributes depending on the input."""
         if None not in list(values.values()):
             return values
 
-        if not values["id"] and None in [values["name"], values["model_uuid"]]:
+        if not values.get("id") and None in [values.get("name"), values.get("model_uuid")]:
             raise ValueError("'id' or 'name and model_uuid' must be set.")
 
-        if values["id"]:
+        if values.get("id"):
```
Mehdi-Bendriss: the .get is not necessary, as you're already safeguarding against it in the previous condition

opensearch_single_kernel/core/models.py
```diff
@@ -188,13 +186,14 @@ class PeerClusterConfig(Model):
...
-    @root_validator
+    @model_validator(mode="before")
+    @classmethod
     def set_node_temperature(cls, values):  # noqa: N805
         """Set and validate the node temperature."""
         allowed_temps = ["hot", "warm", "cold", "frozen", "content"]
 
         input_temps = set()
-        for role in values["roles"]:
+        for role in values.get("roles", []):
```
Mehdi-Bendriss: Could you revisit this (and in other occurrences)? can this ever be None or missing? My worry is that we suddenly start using the `[]` and persist it in opensearch.yml, which effectively means the `coordinating` role.

opensearch_single_kernel/core/models.py
```diff
@@ -188,13 +186,14 @@ class PeerClusterConfig(Model):
     # profile called "testing".
     data_temperature: str | None = None
 
-    @root_validator
+    @model_validator(mode="before")
```
Mehdi-Bendriss: If I understand correctly, this is in preparation of pydantic v2, in which case I believe the equivalent of the v1 `root_validator` is the `@model_validator(mode="after")` not `before` (applies to all occurrences)

opensearch_single_kernel/core/secrets.py
```diff
@@ -34,64 +30,20 @@
...
-
     def label(self, scope: Scope, key: str) -> str:
         """Generated keys to be used within relation data to refer to secret IDs."""
         components = [self.charm.app.name, scope.val]
         if scope == Scope.UNIT:
             components.append(str(self.charm.state.server.unit_id))
         components.append(key)
-        return self.LABEL_SEPARATOR.join(components)
-
-    def breakdown_label(self, label: str) -> dict[str, Any]:
```
Mehdi-Bendriss: where was this code block relocated?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -136,21 +141,22 @@ def _on_peer_relation_joined(self, event: RelationJoinedEvent):
...
         """Handle peer relation changes."""
-        if self.charm.cluster_manager.opensearch_client.is_node_up():
+        # check requirements
+        if not self.charm.state.application.deployment_desc:
+            logger.debug("Deployment description not yet computed.")
+            return
+
+        is_node_up = self.charm.cluster_manager.opensearch_client.is_node_up()
+
+        if is_node_up:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: i'm not sure, as we enrich those in large deployments in the pcluster relation consumer - let's park this one aside for the time being until we complete large deployments and see what needs to be cleaned up

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: sounds good, thank you

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: question still valid

opensearch_single_kernel/managers/cluster.py
```diff
@@ -193,20 +180,23 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
...
         logger.debug(
             "Found deployment description using existing cluster setup. deployment desc: %s",
-            prev_deployment,
+            prev_deployment_desc,
         )
-        directives = prev_deployment.pending_directives
-        deployment_state = prev_deployment.state
+        deployment_desc = (
+            prev_deployment_desc.copy()
+        )  # avoid mutating the previous deployment description in state
```
Mehdi-Bendriss: 1. nit could you add the comment before the line and just have the assignment in a single line?
2. I don't get the idea, and don't think it's needed either, since we retrieve the object from the state (peer relation data) we are effectively already working with a copy. Or do we have a relation data setter for each field? (I also think we're losing some value by omitting the `prev_` prefix)

opensearch_single_kernel/managers/cluster.py
```diff
@@ -387,12 +377,12 @@ def check_if_can_start(self) -> bool:
...
         return False
 
-    def _put_security_index_initialised(self):
+    def _put_security_index_initialised(self) -> None:
         """Set the security index initialized flag."""
         # TODO: Add peer cluster updates here we need to update relations
         self.state.application.is_security_index_initialised = True
 
-    def wait_for_opensearch_up(self):
+    def wait_for_opensearch_up(self) -> None:
```
Mehdi-Bendriss, line 383: thank you for the type hints

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -30,18 +30,29 @@ def __init__(self, state: ClusterState, workload: BaseWorkload):
         self.name = "exclusions_manager"
 
     def _fetch_voting(self) -> set[str]:
```
Mehdi-Bendriss: nit, could you add the suffix `exclusions` to all these methods?

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -58,7 +69,6 @@ def _delete_voting(self, exclusions: set[str], scope: Scope) -> bool:
         # and allows any node to return to the voting config in the future
         try:
             if not self.opensearch_client.remove_exclusions(
-                exclusions=exclusions,
```
Mehdi-Bendriss: is this no longer needed?

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -114,10 +133,18 @@ def add_current(
...
+                    node=node, alt_hosts=self.alt_hosts
+                )
+            except OpenSearchHttpError as e:
+                logger.error(f"Failed to add shard allocation exclusion: {node.name}. Error: {e}")
+                success = False
+            finally:
+                if not success:
+                    logger.error(f"Failed to add shard allocation exclusion: {node.name}.")
+                    if raise_error:
+                        raise OpenSearchExclusionsException("Failed to add allocation exclusion.")
```
Mehdi-Bendriss, line 57: is this caught somewhere?

Mehdi-Bendriss (commented):
Thank you Akram! Great work! 
All that's left for me are small comments and nits and we should be good to

opensearch_single_kernel/common/client.py
akram09: The wait_strategy can be different for each call. We were previously using `wait_exponential` since it will increase the wait time exponentially. It will retry 3 times in both cases but the wait_strategy defines how much time we wait between each retry. I prefer to keep it exponential to give some time to opensearch to heal in case.

opensearch_single_kernel/common/client.py
akram09: Alright :+1:

akram09: Issue can be tracked here #27

opensearch_single_kernel/common/client.py
akram09: I answered this above

opensearch_single_kernel/core/models.py
akram09: With Pydantic v2 it seems that this can not work since it don't have the `__custom_root_type` I believe it is implemented in a different way using the `RootModel` providing `root` propetry directly. Check the PeerClusterFleetApps

opensearch_single_kernel/core/models.py
akram09: Right

opensearch_single_kernel/core/models.py
akram09: That's an interesting knowledge, I will revisit this.

opensearch_single_kernel/core/models.py
akram09: Fixed now :+1:

opensearch_single_kernel/events/opensearch.py
akram09: Fixed, and btw @Mehdi-Bendriss this is one of the instances where I had to go through the cluster_manager to get to the opensearch_client.

opensearch_single_kernel/events/opensearch.py
akram09: Tracked here #28

opensearch_single_kernel/events/opensearch.py
akram09: Sorry, I did test it and it seems okey :+1:

opensearch_single_kernel/managers/cluster.py
akram09: Fixed

akram09: We don't have a data setter for  the fields and we are returning a new deployment_desc anyway

opensearch_single_kernel/managers/exclusions.py
akram09: Why? They will be using as `exclusion_manager.func_name` like `exclusion_manager.fetch_voting` seems logical to me

opensearch_single_kernel/managers/exclusions.py
akram09: yes, it was an omit since we were using it only for logging

opensearch_single_kernel/managers/exclusions.py
akram09: Fixed

opensearch_single_kernel/events/opensearch.py
reneradoi: If I remember correctly, Pedro patched a lot of the code with checks for the deployment description as kind of a "default" sanity check because of a bug he was investigating back then. I think it is a good idea to revisit each occurrence of this condition and check if it makes sense at all.

opensearch_single_kernel/core/models.py
reneradoi: I think it should be `None` instead, to avoid such kind of side effects.

opensearch_single_kernel/charms/base.py
```diff
@@ -58,6 +64,53 @@ def __init__(self, *args):
...
+                if len(nodes) > 1:
+                    # 1. Add current node to the voting + alloc exclusions
+                    self.exclusions_manager.add_current(
+                        node=self.config_manager.current_node,
+                        scope=Scope.APP if self.unit.is_leader() else Scope.UNIT,
+                        voting=True,
+                        allocation=not restart,
+                    )
+            except OpenSearchHttpError:
+                logger.debug("Failed to get online nodes, voting and alloc exclusions not added")
```
reneradoi: Errors should be logged with log level `error` to make them visible to users.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -74,17 +90,216 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+        self.framework.observe(
+            self.charm.on[PEER_RELATION].relation_created,
+            self._on_peer_relation_created,
+        )
+        self.framework.observe(
+            self.charm.on[PEER_RELATION].relation_joined, self._on_peer_relation_joined
+        )
+        self.framework.observe(
+            self.charm.on[PEER_RELATION].relation_changed,
+            self._on_peer_relation_changed,
```
reneradoi, line 108: I'm confused: What is the purpose of this change?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -74,17 +90,216 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+        self.charm.exclusions_manager.add_to_cleanup_list(
+            unit_name=format_unit_name(event.departing_unit.name, deployment_desc.app),
+            scope=Scope.APP if self.charm.unit.is_leader() else Scope.UNIT,
+        )
+
+    def _on_opensearch_data_storage_detaching(
+        self, event: StorageDetachingEvent
+    ) -> None:  # noqa: C901
+        """Triggered when removing unit, Prior to the storage being detached."""
+        # TODO: Warning in case of upgrade in progress
```
reneradoi, line 253: Congrats! By adding `cluster_manager.reconcile_before_unit_removal`, you can now get rid of the `noqa` comment. :)

opensearch_single_kernel/events/opensearch.py
```diff
@@ -74,17 +90,216 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+        planned_units = self.charm.app.planned_units()
+
+        # acquire lock to ensure only 1 unit removed at a time
+        # Closes canonical/opensearch-operator#378
+        if planned_units > 1 and not self.charm.lock_manager.acquired:
+            # Raise uncaught exception to prevent Juju from removing unit
+            raise Exception("Unable to acquire lock: Another unit is starting or stopping.")
+
+        # if the leader is departing, and this hook fails "leader elected" won"t trigger,
```
reneradoi: There's a typo in `won't`.

reneradoi (commented):
Just minor comments and nits, nothing blocking. Good work!

opensearch_single_kernel/events/opensearch.py
akram09: Adding peer relation changed events handler

opensearch_single_kernel/events/opensearch.py
akram09: Thank you !

reneradoi (dismissed):

opensearch_single_kernel/managers/cluster.py
```diff
@@ -162,7 +179,114 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
...
+    def _existing_cluster_setup(
+        self, config: PeerClusterConfig, prev_deployment_desc: DeploymentDescription
+    ) -> DeploymentDescription:
+        """Build deployment description of an existing (started or not) cluster."""
+        logger.debug(
+            "Found deployment description using existing cluster setup. deployment desc: %s",
+            prev_deployment_desc,
+        )
+        # avoid mutating the previous deployment description in state
+        directives = prev_deployment_desc.pending_directives
```
gatici: This still mutates the `prev_deployment_desc.pending_directives`.
We need to do a shallow copy instead of creating a reference such as 
For example: directives = list(prev_deployment_desc.pending_directives)

gatici (dismissed):
Great Work @akram09, thank you!

opensearch_single_kernel/common/client.py
Mehdi-Bendriss: I know what the wait strategy / exponential does, my point is: why not used _always_ use the `wait_fixed(1)` strategy + a retry of 3 times + a timeout of 5 seconds (when host unresponsive)? it's plenty already, why further delay it with the exponential wait and allow this to be configured from call to call?

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: yeah I see where your point about relocating it was coming from

Mehdi-Bendriss (dismissed):
Thanks Akram! please take note on the retry strategy comment for a next PR

Merge pending CI to pass

reneradoi (dismissed):
Re-approval

opensearch_single_kernel/common/client.py
```diff
@@ -165,7 +165,7 @@ def add_allocation_exclusions(
     ) -> bool:
         """Register new allocation exclusions."""
         existing = set() if override else self.fetch_allocation_exclusions(alt_hosts=alt_hosts)
-        all_exclusions = existing.union(allocations or {node.name})
+        all_exclusions = existing.union(allocations if allocations is not None else {node.name})
```
Mehdi-Bendriss, line 168: the previous code was perfectly valid

opensearch_single_kernel/managers/cluster.py
```diff
@@ -188,7 +188,7 @@ def _existing_cluster_setup(
             prev_deployment_desc,
         )
         # avoid mutating the previous deployment description in state
-        directives = prev_deployment_desc.pending_directives
+        directives = list(prev_deployment_desc.pending_directives)
```
Mehdi-Bendriss, line 191: why are we re-casting a list to a list?

opensearch_single_kernel/workload/vm.py
```diff
@@ -122,6 +122,13 @@ def get_host_public_ip(self) -> str | None:
 
         return output.out.strip()
 
+    def is_started(self) -> bool:
+        """Check if OpenSearch is started."""
+        if not (reachable := self.is_reachable(self.host, self.port)):
+            logger.debug("Cannot connect to the OpenSearch server...")
+
+        return reachable
```
s-mlnk: **todo**: function that isn't defined in base class nor used anywhere.

opensearch_single_kernel/common/client.py
akram09: I was havjng issues when opensearch start the charm was putting the node into the allocation exclusions. It was mainly because I hade allocation or {node} whuch translates to a set union

opensearch_single_kernel/managers/cluster.py
akram09: This address gulsum's [comment ](https://github.com/canonical/opensearch-single-kernel-library/pull/16#discussion_r2859658483)

opensearch_single_kernel/managers/config.py
```diff
@@ -306,3 +317,78 @@ def set_jvm_heap_size(self, heap_size_in_kb: int) -> None:
...
+            nodes = self._nodes(
+                use_localhost=self.opensearch_client.is_node_up(), hosts=self.alt_hosts
+            )
+            # update (append) CM IPs
+            self.add_seed_hosts([node.ip for node in nodes if node.is_cm_eligible()])
+        except OpenSearchHttpError:
+            return
+
+    @cached_property
+    def current_node(self) -> Node:  # noqa: C901
```
s-mlnk, line 334: **question**: why can't we use just the API and state's deployment_desc? If so, then it wouldn't be a good match for the ConfigManager. I mean those config file options you get are fully equal to the deployment_desc besides config update latency and bootstrapping process. Or is it common to have invalid deployment_desc with valid config file (that used previous deployment_desc)?

opensearch_single_kernel/core/state.py
s-mlnk, line 1: **question**: why do we use properties twins for the app & unit scopes and then access them through `app_state.property if leader else unit_state.property` pattern? Not the first time I see it. Can't be all put into unit state, especially if app value is used only on the leader and not observed from subordinates (might be wrong here, correct me if so)?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -446,7 +670,8 @@ def _on_start_opensearch(self, event: StartOpenSearch) -> None:  # noqa: C901
             # self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
             return
 
-        self.charm.state.server.update({"started": None})
+        if self.charm.state.server.started:
+            self.charm.state.server.update({"started": ""})
```
s-mlnk, line 675: **suggestion**: it can be updated through properties, however setter is needed to be implemented.

opensearch_single_kernel/managers/cluster.py
```diff
@@ -85,18 +88,32 @@ def _is_connected():
...
+        # update cluster deployment desc
+        logger.debug("Existing deployment_desc before cluster setup: %s", current_deployment_desc)
+        deployment_desc = self._existing_cluster_setup(user_config, current_deployment_desc)
+        logger.debug("Existing deployment_desc after cluster setup: %s", deployment_desc)
+        if current_deployment_desc == deployment_desc:
+            return False
 
-        # TODO: Handle existing cluster case
+        # TODO: Should we add an entry on DeploymentDesc "errors" to reflect on status?
+        self.state.application.put_object("deployment-description", deployment_desc.to_dict())
```
s-mlnk: **suggestion**: not critical, but maybe deployment_desc setter would be a great fit here and in other usages?

opensearch_single_kernel/managers/cluster.py
```diff
@@ -174,11 +298,11 @@ def _user_config(self):
             ],
         )
 
-    def update_bootstrap_state(self, cleanup_application: bool = False):
+    def update_bootstrap_state(self, cleanup_application: bool = False) -> None:
         """Clean up bootstrap state and remove initial_cluster_manager_nodes from config"""
         if cleanup_application:
             self.state.application.update({"bootstrapped": "True"})
-        self.state.server.update({"bootstrap_contributor": None})
+        self.state.server.update({"bootstrap_contributor": ""})
```
s-mlnk, line 305: **suggestion**: can also be moved to setter.

opensearch_single_kernel/managers/cluster.py
```diff
@@ -445,6 +569,53 @@ def recompute_nodes_conf(self, app_id: str, nodes: list[Node]) -> dict[str, Node
...
+            ]
+            self.compute_and_broadcast_updated_topology(remaining_nodes)
+        elif is_last_unit:
+            self.cleanup_on_last_unit_removal()
+
+    def cleanup_on_last_unit_removal(self) -> None:
+        """Clean up cluster state on last unit removal."""
+        if self.state.peer_relation:
+            self.state.application.update({"bootstrap_contributors_count": ""})
+            self.state.application.update({"nodes_config": ""})
```
s-mlnk: **suggestion**: using properties better for sure, but it also can be merged along just like
```python
self.state.application.update({
    "bootstrap_contributors_count": "",
    "nodes_config": "",
})
```

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -4,21 +4,262 @@
...
+from opensearch_single_kernel.utils.helpers import format_unit_name
 from opensearch_single_kernel.workload.base import BaseWorkload
 
+logger = logging.getLogger(__name__)
+
 
 class NodesExclusionsManager(BaseManager):
     """OpenSearch Nodes Exclusions Manager."""
 
+    CONFIG_YML = "opensearch.yml"
```
s-mlnk: **todo**: there's no usages of it.

last_written_value
s-mlnk, line 1: **todo**: and also needs to be deleted:)

s-mlnk (dismissed):
LGTM overall!
Have some comments, but they're not critical.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -74,17 +90,214 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
+            logger.debug("Deployment description not yet computed.")
+            return
+
+        if is_node_up := self.charm.cluster_manager.opensearch_client.is_node_up():
+            health = self.charm.status.apply_health(app=self.charm.unit.is_leader())
+
+            if health in [HealthColors.UNKNOWN, HealthColors.YELLOW_TEMP]:
+                # we defer because we want the temporary status to be updated
+                logger.debug("Cluster health temp yellow or unknown. Deferring event.")
+                event.defer()
```
s-mlnk, line 157: todo: missing `return`.

opensearch_single_kernel/managers/config.py
akram09: I believe we can do that, the idea is to have the API as source of truth and if the API is not available we are moving towards using the state as the source of truth now especially with your PR

opensearch_single_kernel/core/state.py
akram09: The thing is that we don't want to use the property is_leader since it is charm specific and we want to have the state not relying on that.

opensearch_single_kernel/events/opensearch.py
akram09: This will actually delete the property not set it to a certain value

Mehdi-Bendriss (dismissed):

opensearch_single_kernel/managers/cluster.py
Mehdi-Bendriss: from the code I don't see it having an impact

opensearch_single_kernel/managers/config.py
s-mlnk: Should I do this or you?

opensearch_single_kernel/core/state.py
s-mlnk: I mean could we use only unit state? This way we won't need to differentiate between 2 state objects with is_leader.

opensearch_single_kernel/events/opensearch.py
s-mlnk: Setter can accept None.

opensearch_single_kernel/managers/cluster.py
gatici: In the following lines, directives are mutated using append and remove as below. The list call creates a new list, hence mutations won't modify `prev_deployment_desc.pending_directives` in-place.  That's why we need this change.
```
 directives.append(Directive.SHOW_STATUS)
 directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
 ```

gatici (dismissed):

opensearch_single_kernel/managers/config.py
akram09: I believe we should do it on your PR.

opensearch_single_kernel/core/state.py
akram09: We don't want to change the design, since it's a refactoring of the charm

opensearch_single_kernel/managers/cluster.py
akram09: Fixed

Mehdi-Bendriss (approved):

gatici (approved):

PR #17, opened 2026-01-30
patch: [DPE-8939] Refactor config manager

s-mlnk:
# Key changes

## `opensearch.yml` config refactor

Currently, there are many functions for updating & cleaning specific parts of the `opensearch.yml` config all across the codebase. It results in spaghetti-like config manipulation that is hard to keep track of. This PR refactors it into the single "generate config" function that will improve code readability, will be less prone to development errors and will simplify further development.

## `unicast_hosts.txt` aka `seed hosts` config refactor

Currently, seed hosts are set both from state's `nodes_config` and directly from API in chaotic, unorganised way. This PR refactors it to primarily use `nodes_config` source. The only exception is when `update_opensearch_config` is fed with overridden bootstrap-specific `roles` & `cm_names` & `cm_ips` values.

## Static config options refactor

Currently, static config options are set both along with "dynamic" values (e.g. `ConfigManager.set_node` updates both dynamic and static options) and from distinct "set static options" functions (e.g. `ConfigManager.set_client_auth`). This PR moves all of the "set static options" operations into separate functions and moves their calls to the single place - `ConfigManager.update_opensearch_config`. This removes event handlers responsibility of updating those and improves codebase readability.


opensearch_single_kernel/charms/base.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, *args):
         self.lock_manager = LockManager(self.state, self.workload)
         self.profiles_manager = ProfilesManager(self.state, self.workload)
         self.health_manager = HealthManager(self.state, self.workload)
-        self.config_manager = ConfigManager(self.state, self.workload)
+        self.config_manager = ConfigManager(self.state, self.workload, self.cluster_manager)
```
akram09: It is better to avoid having a manager depends on another manager

opensearch_single_kernel/events/opensearch.py
```diff
@@ -163,12 +164,10 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
         # for relation in self.model.relations.get(ClientRelationName, []):
         # self.opensearch_provider.update_endpoints(relation)
 
-        # register new cm addresses on every node
-        self.charm.config_manager.add_cm_addresses_to_conf()
```
akram09, line 167: Why are we removing this ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -416,7 +415,8 @@ def _on_install(self, event: InstallEvent) -> None:
 
     def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """On config changed event. Useful for IP changes or for user provided config changes."""
-        if self.charm.config_manager.update_host_if_needed():
+        # TODO: only on IP change
```
akram09, line 418: What are you planning to do here I see a TODO

opensearch_single_kernel/events/opensearch.py
```diff
@@ -729,11 +722,17 @@ def _on_start_opensearch(self, event: StartOpenSearch) -> None:  # noqa: C901
...
+            computed_roles = self.charm.state.computed_roles()
+            cm_names = self.charm.cluster_manager.get_cluster_managers_names(nodes)
+            cm_ips = self.charm.cluster_manager.get_cluster_managers_ips(nodes)
+            self.charm.cluster_manager.configure_bootstrap_contributors(
+                computed_roles, cm_names, cm_ips
+            )
 
-            # Set the configuration of the node
-            self._set_node_conf(nodes)
+            self.charm.config_manager.update_opensearch_config(cm_names=cm_names, cm_ips=cm_ips)
```
akram09, line 735: nice

opensearch_single_kernel/events/opensearch.py
```diff
@@ -163,12 +164,10 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
         # for relation in self.model.relations.get(ClientRelationName, []):
         # self.opensearch_provider.update_endpoints(relation)
 
-        # register new cm addresses on every node
-        self.charm.config_manager.add_cm_addresses_to_conf()
-
         if self.charm.unit.is_leader():
             # Recompute the node roles in case self-healing didn't trigger leader related event
             self.charm.cluster_manager.recompute_roles_if_needed()
+            self.charm.config_manager.update_nodes_config()
```
akram09, line 170: Why we are updating nodes config on leader ?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +41,287 @@ class ConfigManager(BaseManager):
...
+        cluster_manager: ClusterManager,
+    ):
+        self.state = state
+        self.workload = workload
+        self.cluster_manager = cluster_manager
 
     @property
-    def yaml_setter(self):
-        """Return the yaml_setter."""
+    def yaml_setter(self) -> YamlConfigSetter:
```
akram09, line 52: Thanks for this

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +41,287 @@ class ConfigManager(BaseManager):
...
+        return (
+            {"node.attr.temp": self._opensearch_data_temperature}
+            if self._opensearch_data_temperature
+            else {}
         )
 
-    def cleanup_initial_cluster_managers(self) -> None:
-        """Update the opensearch.yaml by deleting initiali_cluster_manager_nodes."""
-        self.yaml_setter.delete(self.CONFIG_YML, "cluster.initial_cluster_manager_nodes")
+    def _opensearch_manager_config(
```
akram09: nit: `opensearch_cluster_manager_config`

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +41,287 @@ class ConfigManager(BaseManager):
...
-                "PKCS12",
-            )
-
-            self.yaml_setter.put(
-                self.CONFIG_YML,
-                f"plugins.security.ssl.{target_conf_layer}.{store_type}_filepath",
-                f"{self.workload.paths.certs_relative}/{cert if cert == 'ca' else cert_type.val}.p12",
-            )
-
+    def _update_static_security_options(self):
```
akram09, line 277: Why are we using the yaml setter here ? can't this be included in the reconcile function ?

akram09 (commented):
Thanks @s-mlnk, some comments from me. I will do some manual testing in the meantime

opensearch_single_kernel/events/opensearch.py
```diff
@@ -416,7 +415,8 @@ def _on_install(self, event: InstallEvent) -> None:
 
     def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """On config changed event. Useful for IP changes or for user provided config changes."""
-        if self.charm.config_manager.update_host_if_needed():
+        # TODO: only on IP change
+        if self.charm.config_manager.update_opensearch_config():
```
akram09, line 419: Seems that this is returning True when the charm is started this should return True when updating the host not when setting a new value. This is what `update_host_if_needed` was doing

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +41,287 @@ class ConfigManager(BaseManager):
...
-        self.yaml_setter.put(self.CONFIG_YML, "plugins.security.disabled", False)
-        self.yaml_setter.put(self.CONFIG_YML, "plugins.security.ssl.http.enabled", True)
-        self.yaml_setter.put(
-            self.CONFIG_YML,
-            "plugins.security.ssl.transport.enforce_hostname_verification",
-            True,
-        )
+        if (
+            self.workload.paths.opensearch_config.exists()
+            and self.workload.paths.opensearch_config.read_text() == content
```
akram09, line 99: nit question for @Mehdi-Bendriss : Do we need the comments ? Since with this approach we are overriding the comments.But I don't think we have the same problem of ruaml throwing exceptions.

akram09, line 99: Also, we need to make sure the lists are sorted so that we don't have unnecessary writes

opensearch_single_kernel/charms/base.py
s-mlnk: I accidentally left that :)

opensearch_single_kernel/events/opensearch.py
s-mlnk: Because I've added `self.charm.config_manager.update_nodes_config()` that does the same thing below. And if not a leader, `self.charm.config_manager.update_opensearch_config()` will include former command.

opensearch_single_kernel/events/opensearch.py
s-mlnk: It's not a full config update, it's `self.charm.config_manager.add_cm_addresses_to_conf()` alternative from above.

opensearch_single_kernel/events/opensearch.py
s-mlnk: Would it be better to write IP address to the state and then compare those values? I'm just not a particular fan of reading back the config.
@Mehdi-Bendriss

opensearch_single_kernel/managers/config.py
s-mlnk: Reconcile function done only for the `opensearch.yml` file.
I propose to refactor security config (into reconcile method) later along with JWT or OIDC PR.

opensearch_single_kernel/events/opensearch.py
s-mlnk: Also I confirmed the integration tests failures are caused by this specific change :)

Mehdi-Bendriss: I don't particularly feel strongly about reading the config file as a source of truth, as a matter of fact by doing so we always have the risk of reading a file that hasn't been reloaded yet. 
Which means that our _real_ source of truth is the `_nodes or _cluster` APIs to see what the cluster actually sees. 
But in the case where the node is down due to a network partition for ex, and/or cleared from the reported output of those APIs, we can base our logic on the state

PR #18, opened 2026-02-05
patch: Refactor OpenSearch provider relation

akram09:
This Pull Request integrates the refactoring of OpenSearch provider. This includes event handling of `index-created`, `relation-changed`, `relation-broken`, and `relation-depated`. 

The PR introduces a new state model `ExternalOpenSearchClient` which simplifies the interaction with the relation databag. The events handling is done using `ExternalCliientsEventsHandler` and the manager `ExternalClientsManager`.  The following diagrams illustrates the events handling processes: 

image
image
image
image


last_written_value
```diff
@@ -0,0 +1 @@
+2117
```
reneradoi: This file should not be committed.

opensearch_single_kernel/managers/internal_users.py
```diff
@@ -0,0 +1,148 @@
...
+
+    This manager handles everything related to configuring users in OpenSearch.
+    """
+
+    def __init__(self, state: ClusterState, workload: BaseWorkload):
+        super().__init__(state, workload)
+        self.name = "internal_users_manager"
+        self.yaml_setter = YamlConfigSetter(self.workload)
+
+    def put_or_update_internal_user_leader(
```
reneradoi, line 46: I know the method name was taken over, but it is a bit confusing given that it is also run by non-leaders.

opensearch_single_kernel/managers/internal_users.py
```diff
@@ -0,0 +1,148 @@
...
+        if user in OPENSEARCH_SYSTEM_USERS:
+            self.state.secrets.put(Scope.APP, self.state.secrets.hash_key(user), hashed_pwd)
+
+        if user == ADMIN_USER:
+            self.state.application.update({"admin_user_initialized": "True"})
+
+    def _put_or_update_internal_user_unit(self, user: str) -> None:
+        """Create system user or update it with a new password."""
+        # Leader is to set new password and hash, others populate existing hash locally
+        hashed_pwd = self.state.secrets.get(Scope.APP, self.state.secrets.hash_key(user))
```
reneradoi: Comments here are identical with the previous method, though they do different things.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -790,7 +1046,55 @@ def _on_secret_changed(self, event: SecretChangedEvent) -> None:  # noqa: C901
...
+        logger.debug("Secret change for %s", str(label_key))
+
+        if is_leader and label_key == self.charm.state.secrets.password_key(KIBANA_SERVER_USER):
+            self.charm.external_clients_manager.update_dashboards_password()
+
+        # Non-leader units need to maintain local users in internal_users.yml
+        elif not is_leader and label_key in system_user_hash_keys:
+            password = event.secret.get_content()[label_key]
+            if sys_user := self.charm.state.secrets._user_from_hash_key(label_key):
+                self.charm.internal_users_manager.put_internal_user(sys_user, password)
```
reneradoi: We shouldn't access protected methods. Why not using `internal_users_manager.save_user_locally()`? It seems to be intended for this case.

opensearch_single_kernel/managers/internal_users.py
```diff
@@ -0,0 +1,148 @@
...
+                self.yaml_setter.delete("opensearch-security/internal_users.yml", user)
+
+    def save_user_locally(self, user: str):
+        """Save the user in internal_users.yaml"""
+        user_hash = self.state.secrets.hash_key(user)
+        hashed_pwd = self.state.secrets.get(Scope.APP, user_hash)
+        # System users have to be saved locally in internal_users.yml
+        self.put_internal_user(user, hashed_pwd)
+
+    def put_internal_user(self, user: str, hashed_pwd: str):
```
reneradoi: This method could make use of a switch case statement for better readability.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+    from opensearch_single_kernel.charms.base import OpenSearchBaseCharm
+
+logger = logging.getLogger(__name__)
+
+
+class ExternalClientsEventsHandler(Object):
+    """Handler for Charm External Clients Events."""
+
+    def __init__(self, charm: "OpenSearchBaseCharm") -> None:
+        super().__init__(charm, key="opensearch_events")
```
reneradoi: This needs to be updated to resolve a duplicate name.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.cluster_manager.opensearch_client.is_node_up() or not event.index:
+            event.defer()
+            return
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
+            logger.error("No external client found for relation id %d", event.relation.id)
```
reneradoi: You can combine these two lines with the `:=` operator

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+            event.defer()
```
reneradoi: Does it make sense to defer in this case? How could it happen, and how would deferring and coming back solve it?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+        self.charm.external_clients_manager.update_relation_index(external_client, event.index)
+        self.charm.external_clients_manager.update_relation_tls_info(external_client)
+        # Get nodes
+        try:
+            nodes = self.charm.cluster_manager.get_nodes(use_localhost=True)
+        except OpenSearchHttpError:
+            logger.error("unable to get nodes")
+            nodes = []
+
+        self.charm.external_clients_manager.update_relation_endpoints(external_client, nodes)
```
reneradoi: Is it acceptable from external client side that we do not put the endpoints into the relation data? I think it will be a very rare case that we are able to create an index (some lines above) but are not able to get the cluster nodes. But in case it happens, wouldn't it make sense to have proper error handling here instead of just moving on?

reneradoi: Another comment here: I'm confused, why don't we use `self.update_external_client_endpoints()` here?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+            CharmStatuses.NEW_INDEX_REQUESTED, pattern=Status.CheckPattern.Interpolated
+        )
+        self.charm.status.clear(
+            CharmStatuses.INDEX_CREATION_FAILED, pattern=Status.CheckPattern.Interpolated
+        )
+        self.charm.status.clear(
+            CharmStatuses.USER_CREATION_FAILED, pattern=Status.CheckPattern.Interpolated
+        )
+
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
```
reneradoi, line 147: Docstring is missing here.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+        if not self.charm.unit.is_leader():
+            return
+
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
+            logger.error("No external client found for relation id %d", event.relation.id)
+            event.defer()
+            return
+        if self.charm.cluster_manager.opensearch_client.is_node_up():
```
reneradoi: Same question here as on the `index_requested` handler: Does it make sense to defer in this case? How could it happen, and how would deferring and coming back solve it?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+            event.defer()
+
+    def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
+        """Check if this relation is being removed, and update the peer databag accordingly."""
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
+            logger.error("No external client found for relation id %d", event.relation.id)
+            event.defer()
+            return
+        # remove departing unit from endpoints available to requirer charm.
```
reneradoi: Same question on the deferral: Does it make sense to defer in this case? How could it happen, and how would deferring and coming back solve it?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+            event.defer()
+            return
+        if self.charm.cluster_manager.opensearch_client.is_node_up():
+            self.update_external_client_endpoints(external_client)
+        else:
+            event.defer()
+
+    def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
+        """Check if this relation is being removed, and update the peer databag accordingly."""
+        external_client = self.charm.state.external_client_by_relation(event.relation)
```
reneradoi, line 165: Please be aware that for relation-departed and relation-broken events, relation data (and all other state information) of the unit may already be gone and for example this call will not return relation data anymore (see https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.RelationDepartedEvent). Especially relevant considering the following deferral on the next lines.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+        """Handle client relation-broken event."""
+        if not self.charm.unit.is_leader():
+            return
+
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
+            logger.error("No external client found for relation id %d", event.relation.id)
+            event.defer()
+            return
+        if external_client.is_relation_departing():
```
reneradoi: Same comment here: You will most probably not have relation data anymore for a relation that is gone. This could result in a loop of deferred events.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1219,8 +1222,15 @@ def request_new_admin_certificate(self) -> None:
 
         self.charm.tls_events.certs.request_certificate_creation(certificate_signing_request=csr)
 
-    def on_unit_ip_changed(self, event: ConfigChangedEvent) -> None:
-        """Triggered when the unit IP is changed."""
-        self.charm.status.set(CharmStatuses.TLS_NEW_CERTS_REQUESTED)
-        self.charm.tls_manager.delete_stored_tls_resources()
-        self.request_new_unit_certificates()
+    def update_external_clients_endpoints(self) -> None:
```
reneradoi, line 1169: We have duplicate code here, as the same method is in the `ExternalClientsEventsHandler` class. I would recommend to change the code to the following and use it directly in the event handler methods, avoiding this "wrapper" method in both event handler classes (and potentially some more in the future, think large deployments):

```python
if not self.charm.unit.is_leader():
    return

try:
    nodes = self.charm.cluster_manager.get_nodes(use_localhost=True)
except OpenSearchHttpError:
    <raise or adequate error handling>

self.charm.external_clients_manager.update_relation_endpoints(nodes)
```

The loop over `self.charm.state.external_clients` should be done by the external clients manager, who's responsibility this is instead of multiple event handlers.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+        # if self.charm.upgrade_in_progress:
+        #    logger.warning(
+        # "Modifying relations during an upgrade is not supported."
+        # "The charm may be in a broken, unrecoverable state"
+        # )
+        self.charm.external_clients_manager.remove_lingering_relation_users_and_roles(
+            external_client
+        )
+
+    def update_external_client_endpoints(
```
reneradoi, line 207: Please see my recommendation on the same method in the `OpensearchEventHandler` class on how to refactor this.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -357,21 +355,19 @@ def _on_update_status(self, event: UpdateStatusEvent):  # noqa: C901
             if health == HealthColors.UNKNOWN:
                 return
 
-        # TODO: Handle client relations updates
-        # for relation in self.model.relations.get(ClientRelationName, []):
-        # self.opensearch_provider.update_endpoints(relation)
-
-        # deployment_desc = self.charm.state.application.deployment_desc
+        self.update_external_clients_endpoints()
+        deployment_desc = self.charm.state.application.deployment_desc
```
reneradoi: We already check for the deployment description in line 327, we can simplify the check here and in the following lines.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -414,11 +410,11 @@ def _on_install(self, event: InstallEvent) -> None:
             except OpenSearchInstallError:
                 self.charm.status.set(CharmStatuses.INSTALL_ERROR)
 
-    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:  # noqa: C901
         """On config changed event. Useful for IP changes or for user provided config changes."""
         if self.charm.config_manager.update_host_if_needed():
             # This happens when the unit IP has changed
-            self.on_unit_ip_changed(event)
+            self.charm.tls_events.on_unit_ip_changed(event)
```
reneradoi: Instead of creating dependencies between event handlers (which is not recommended), you can observe for the config changed event in the `TLSEventHandler` class and have the specific logic there.

Example from `charmed-etcd`:
- event handling for config changed in the base events handler: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/events/etcd.py#L234
- TLS related event handling for config changed: https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/events/tls.py#L342

Thereby you can also avoid the disabling of quality checks with `noqa: C901`.

opensearch_single_kernel/events/tls.py
```diff
@@ -317,6 +318,12 @@ def _on_certificate_invalidated(self, event: CertificateInvalidatedEvent) -> Non
         logger.debug(f"Received certificate invalidation. Reason: {event.reason}")
         self._on_certificate_expiring(event)
 
+    def on_unit_ip_changed(self, event: ConfigChangedEvent) -> None:
```
reneradoi: As commented in the `OpensearchEventHandler` class: Rather listen to the `config_changed` event here, too, and perform TLS related operations here instead of calling one event handler from the other.

opensearch_single_kernel/events/tls.py
```diff
@@ -317,6 +318,12 @@ def _on_certificate_invalidated(self, event: CertificateInvalidatedEvent) -> Non
...
+        """Triggered when the unit IP is changed."""
+        self.charm.status.set(CharmStatuses.TLS_NEW_CERTS_REQUESTED)
+        self.charm.tls_manager.delete_stored_tls_resources()
+        self.request_new_unit_certificates()
```
reneradoi: This method doesn't exist.

opensearch_single_kernel/managers/base.py
```diff
@@ -102,3 +106,17 @@ def _nodes(
                     )
                     nodes.append(node)
         return nodes
+
+    @cached_property
+    def version(self) -> str:
```
reneradoi: Is this planned to be used in other managers than the `ExternalClientsManager`? If not, I would move it out of the base class.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,271 @@
...
+        # this would create an invalid role config.
+        for role in roles:
+            if role.upper() in ExtraUserRolePermissions._member_names_:
+                for perm_scope, perms in ExtraUserRolePermissions[role.upper()].value.items():
+                    permissions[perm_scope] += perms
+
+        for perm_set in permissions["index_permissions"]:
+            # If this isn't a set of admin permissions (which applies to all indices) then set it
+            # to index.
+            if perm_set["index_patterns"] == []:
```
reneradoi: This could be simplified by just checking if it exists.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,271 @@
...
+    ):
+        """Create a relation user.
+
+        Relation users are registered with a dedicated role which maps to the username,
+        and their name is saved in the databag for later reference.
+        """
+        self.opensearch_client.create_role(role_name=user, permissions=permissions)
+        users = self.state.application.client_users_dict
+
+        if users.get(relation_id):
```
reneradoi: Comparing an `int` with `dict[str, str]`.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,271 @@
...
+            )
+
+        self.opensearch_client.create_user(user, [user], hashed_pwd)
+        self.opensearch_client.create_role_mapping(
+            user, self.state.get_relation_mapped_users(user)
+        )
+        users[str(relation_id)] = user
+        self.state.application.client_users_dict = users
+
+    def update_relation_version(self, external_client: ExternalOpenSearchClient):
```
reneradoi: You could call the state setter directly in the event handler instead of passing around the external client and maintaining these methods (applies to `update_relation_index`and `update_relation_credentials` as well).

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,212 @@
...
+                dynamic_params={"rel_name": CLIENT_RELATION, "id": event.relation.id},
+            )
+            return
+
+        self.charm.external_clients_manager.update_relation_version(external_client)
+        self.charm.external_clients_manager.update_relation_credentials(
+            external_client, username, pwd
+        )
+        self.charm.external_clients_manager.update_relation_index(external_client, event.index)
+        self.charm.external_clients_manager.update_relation_tls_info(external_client)
```
reneradoi: We should have error handling here as this method may raise an error which would leave the charm in irrecoverable state.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,271 @@
...
+        Args:
+            departed_external_client: if a relation is departing, pass in the
+            ExternalOpenSearchClient and its user will be deleted.
+        """
+        if not self.opensearch_client.is_node_up():
+            return
+        relation_users = self.state.application.client_users_dict
+
+        if departed_external_client and (
+            not relation_users or departed_external_client.relation.id not in relation_users
```
reneradoi: Heads up: Given that we call this method on the `relation_departed` and `relation_broken` events, it might happen that the `relation.id` is already gone and you cannot use it anymore.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,271 @@
...
+                try:
+                    self.opensearch_client.remove_role_mapping(username)
+                except OpenSearchUserMgmtError:
+                    logger.error(f"failed to remove role mapping for {username}")
+
+                del relation_users[rel_id]
+        self.state.application.client_users_dict = relation_users
+
+    def update_relations_roles_mapping(self) -> bool:
+        """Updates all the relations roles mapping due to config change.
```
reneradoi, line 238: For a cleaner implementation, you could simplify a bit here (paradigm is: Let the error bubble up):

```python
if not self.opensearch_client.is_node_up():
    raise OpenSearchUserMgmtError

for user in self.state.application.client_users_dict:
    self.opensearch_client.create_role_mapping()
```

And in the event handler:

```python
try:
    self.charm.external_clients_manager.update_relations_roles_mapping()
except (OpenSearchUserMgmtError, OpenSearchUserMgmtError):
    logger.error()
    event.defer()
```

opensearch_single_kernel/managers/tls.py
```diff
@@ -48,6 +48,12 @@ class TlsManager(BaseManager):
     can be used by different events handlers not only tls events handler.
     """
 
+    CA_ALIAS = "ca"
+    OLD_CA_ALIAS = f"old-{CA_ALIAS}"
+    KEYTOOL = "opensearch.keytool"
+    OLD_CA_PREFIX = "old-"
+    CERTS_EXPIRATION_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
```
reneradoi: I think this is an inconsistency because of the PR base, but these literals are also imported some lines above.

opensearch_single_kernel/common/client.py
```diff
@@ -44,6 +52,279 @@ def __init__(
...
+            if (
+                e.response_code == 400
+                and e.response_body.get("error", {}).get("type")
+                == "resource_already_exists_exception"
+            ):
+                logger.warning("Index failed to be created as it already exists, continuing...")
+            else:
+                raise e
+
+    def get_roles(self) -> dict[str, Any]:
```
reneradoi: This method is not used currently, could be removed if not required for other features.

opensearch_single_kernel/common/client.py
```diff
@@ -44,6 +52,279 @@ def __init__(
...
+            else:
+                raise OpenSearchUserMgmtError(e)
+
+        logger.debug(resp)
+        if resp.get("status") != "OK":
+            raise OpenSearchUserMgmtError(f"removing role {role_name} failed")
+
+        return resp
+
+    def get_users(self) -> dict[str, Any]:
```
reneradoi: Same here: method currently not used, can be removed.

opensearch_single_kernel/common/client.py
```diff
@@ -44,6 +52,279 @@ def __init__(
...
+            else:
+                raise OpenSearchUserMgmtError(e)
+
+        logger.debug(resp)
+        if resp.get("status") != "OK":
+            raise OpenSearchUserMgmtError(f"removing user {user_name} failed")
+        return resp
+
+    def patch_user(self, user_name: str, patches: list[dict[str, Any]]) -> dict[str, Any]:
+        """Applies patches to user.
```
reneradoi, line 200: This method may raise, please add error handling to where it is called.

opensearch_single_kernel/common/client.py
```diff
@@ -44,6 +52,279 @@ def __init__(
...
+                    "status": "OK",
+                    "response": "role mapping does not exist, and therefore has not been removed",
+                }
+            else:
+                raise OpenSearchUserMgmtError(e)
+
+        if resp.get("status") != "OK":
+            raise OpenSearchUserMgmtError(f"removing role mapping {role} failed")
+
+    def update_user_password(self, username: str, hashed_pwd: str):
```
reneradoi, line 274: Same here: This may raise, please add error handling to the event handlers that call it (via `InternalUserManager.put_or_update_internal_user_leader()`).

opensearch_single_kernel/common/statuses.py
```diff
@@ -68,6 +68,13 @@ class CharmStatuses(Enum):
     )
     WAITING_TO_START = WaitingStatus("Waiting for OpenSearch to start...")
 
+    # Client
+    NEW_INDEX_REQUESTED = MaintenanceStatus("new index {index} requested")
+    INDEX_CREATION_FAILED = BlockedStatus(
```
reneradoi: Blocked status means `The charm is stuck. Human intervention is required.` ([reference](https://documentation.ubuntu.com/juju/3.6/reference/status/#unit-status)). What would be the human intervention for this case?

opensearch_single_kernel/core/state.py
```diff
@@ -353,6 +358,127 @@ def is_data_role_in_cluster_fleet_apps(self) -> bool:
...
+    def client_users_dict(self) -> dict[str, str]:
+        """Get the client relation users dict from application databag."""
+        users_dict = self.get_object("client_relation_users") or {}
+        return users_dict
+
+    @client_users_dict.setter
+    def client_users_dict(self, users_dict: dict[str, str]):
+        """Set the client relation users dict in application databag."""
+        self.put_object("client_relation_users", users_dict)
+
```
reneradoi, line 380: It feels like these two could be removed and instead the state getter and setter just be called directly (as done in so many other cases).

opensearch_single_kernel/core/state.py
```diff
@@ -353,6 +358,127 @@ def is_data_role_in_cluster_fleet_apps(self) -> bool:
...
+        """Set the relation departing flag in relation databag."""
+        self.update({self.depart_flag: "True"})
+
+    def is_relation_departing(self) -> bool:
+        """Check if the relation departing flag is set in relation databag."""
+        return self.relation_data.get(self.depart_flag, "") == "True"
+
+    def delete_relation_departing_flag(self) -> None:
+        """Delete the relation departing flag from relation databag."""
+        self.update({self.depart_flag: None})
```
reneradoi: Deleting an item from the relation data should happen by updating it to an empty string instead of `None`.

reneradoi (commented):
Hey Akram! Thank you again for cleaning up the PR, that has helped reviewing. I have left a couple of comments and questions for consideration. Could you please also confirm whether the `application charm` for testing and the integration tests have been taken over or should be reviewed?

opensearch_single_kernel/managers/internal_users.py
akram09: I think we don't actually call it  on non-leader units

opensearch_single_kernel/managers/internal_users.py
akram09: I fixed the confusion. It was previously fixed but apparently I forget to remove this function

opensearch_single_kernel/managers/internal_users.py
akram09: I don't think we can do a switch over constant variables ? We could have it technically but we only use the `case _:` in that case . Please correct me if I am wrong

opensearch_single_kernel/events/external_clients.py
akram09: You are right

opensearch_single_kernel/events/external_clients.py
akram09: Updating endpoints will be done in multiple cases, for instance on update_status  and on peer_relation changed. So I think it would be fine since it will be reconciled

akram09: I have duplicated the code indeed. Thank you

opensearch_single_kernel/events/external_clients.py
akram09: Fixed, thank you

opensearch_single_kernel/events/external_clients.py
akram09: Thank you, I removed the event deferral

opensearch_single_kernel/events/opensearch.py
akram09: Updated in openserch event handler. Though in external client event handler I would still keep it.

opensearch_single_kernel/events/external_clients.py
akram09: This is called 4 times in the events handler I would rather have it in a separate function to not have duplicated code in the ExternalClientsEventHandler

opensearch_single_kernel/events/opensearch.py
akram09: Fixed

opensearch_single_kernel/events/opensearch.py
akram09: Yes, I totally Agree

opensearch_single_kernel/managers/external_clients.py
akram09: I didn't get your comment

opensearch_single_kernel/managers/tls.py
akram09: Fixed

opensearch_single_kernel/common/statuses.py
akram09: I didn't design this but I am assuming it would be to remove the integration and integrate again. We can update the message

opensearch_single_kernel/core/state.py
akram09: I personally prefer it this way

akram09:
> Hey Akram! Thank you again for cleaning up the PR, that has helped reviewing. I have left a couple of comments and questions for consideration. Could you please also confirm whether the `application charm` for testing and the integration tests have been taken over or should be reviewed?

To reply on this comment. Indeed, I have moved the integration tests and the application-charm from opensearch-operator repository, no updates there.

opensearch_single_kernel/managers/internal_users.py
reneradoi: Hm, not sure anymore where I found it, maybe this was updated. Anyway, we can resolve this, I have double checked again.

opensearch_single_kernel/managers/internal_users.py
reneradoi: Yes that is true, it would have to be like this:

```
match user:
    case "admin":
        logger.info("something")
    case "kibanaserver":
        logger.info("something else")
    case _:
        raise
```

Up to you, I'm fine with leaving it as is.

opensearch_single_kernel/events/opensearch.py
reneradoi: Yeah it's not a big issue, just wanted to point to it.

opensearch_single_kernel/managers/external_clients.py
reneradoi: `if not perm_set["index_patterns"]:`

opensearch_single_kernel/common/statuses.py
reneradoi: Yes, would probably be more helpful for users.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,302 @@
...
+    def update_relation_credentials(
+        self, external_client: ExternalOpenSearchClient, username: str, password: str
+    ):
+        """Update the relation databag with credentials."""
+        external_client.username = username
+        external_client.password = password
+
+    def update_relation_tls_info(
+        self, external_client: ExternalOpenSearchClient, ca_chain: str | None = None
+    ):
```
reneradoi: This method raises some questions:
- It has `ca_chain` in its signature, but is never called with this parameter.
- It just returns in case of `AttributeError` "presumably we don't have a TLS relation", even though the Opensearch node cannot come up without a TLS relation being set up. Isn't this an issue for a client application if the TLS information is not set correctly to relation data?
- It handles a `TypeError` - when could that happen?

I would recommend to simplify the logic, let the errors bubble up and make it clearer. Actually, I believe it can be removed and in the event handler instead just be called like this:
```
try:
    external_client.tls_ca = self.charm.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)["chain"]
except KeyError:
    event.defer()
```

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,302 @@
...
+        try:
+            self.opensearch_client.create_role_mapping(
+                user, self.state.get_relation_mapped_users(user)
+            )
+        except OpenSearchHttpError as e:
+            raise OpenSearchUserMgmtError(e)
+        users[str(relation_id)] = user
+        self.state.application.client_users_dict = users
+
+    def update_relation_credentials(
```
reneradoi: This method seems to not be used anymore, can be removed.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,302 @@
...
+
+    def update_dashboards_password(self):
+        """Update each Opensearch Dashboards relation with the latest kibanaserver."""
+        pwd = self.state.secrets.get(Scope.APP, password_key(KIBANA_SERVER_USER))
+        for dashboards_client in self.state.dashboards_clients:
+            dashboards_client.username = KIBANA_SERVER_USER
+            dashboards_client.password = pwd
+
+    @cached_property
+    def version(self) -> str:
```
reneradoi, line 265: Seems like the comment of this property is outdated, as this is no longer a GET request but rather a `run_cmd`. Heads up: This might raise, but the error is not caught currently.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,205 @@
...
+            CharmStatuses.USER_CREATION_FAILED, pattern=Status.CheckPattern.Interpolated
+        )
+
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
+        """Handle opensearch client relation-changed event."""
+        if not self.charm.unit.is_leader():
+            return
+
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
```
reneradoi, line 153: That is confusing me: How can this happen? We get a relation changed event for a client relation but don't have the client relation?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,205 @@
...
+    def update_external_client_endpoints(
+        self, external_client: ExternalOpenSearchClient, omit_endpoints: set | None = None
+    ) -> None:
+        """Update the external client state with endpoints."""
+        if self.charm.unit.is_leader():
+            try:
+                nodes = self.charm.cluster_manager.get_nodes(use_localhost=True)
+            except OpenSearchHttpError:
+                logger.error("unable to get nodes")
+                nodes = []
```
reneradoi, line 216: I am afraid this will clear the endpoint list in the relation databag in case of error. Is this intentional?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,205 @@
...
+            # This unit is being removed.
+            external_client.delete_relation_departing_flag()
+            return
+        # TODO: Handle upgrades
+        # if self.charm.upgrade_in_progress:
+        #    logger.warning(
+        # "Modifying relations during an upgrade is not supported."
+        # "The charm may be in a broken, unrecoverable state"
+        # )
+        self.charm.external_clients_manager.remove_lingering_relation_users_and_roles(
```
reneradoi, line 203: This already happens on `relation_departed`.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,205 @@
...
+            logger.error("No external client found for relation id %d", event.relation.id)
+            return
+        # remove departing unit from endpoints available to requirer charm.
+        if event.departing_unit.app == self.charm.app:
+            departing_unit_ip = self.charm.state.unit_ip(event.departing_unit)
+            self.update_external_client_endpoints(
+                external_client, omit_endpoints={departing_unit_ip}
+            )
+        if event.departing_unit == self.charm.unit:
+            external_client.set_relation_departing()
```
reneradoi: So this whole `relation_departing` flag is confusing me: Given that on `relation_departed`, updates to relation data are not published to other units anymore, only this unit can see this flag. And all it does is check the flag on `relation_broken` to avoid calling `remove_lingering_users` in case it is set, but this method is also already called here (two lines below). Am I missing something, or does this not make sense?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,205 @@
...
+        if not external_client:
+            logger.error("No external client found for relation id %d", event.relation.id)
+            event.defer()
+            return
+        if self.charm.cluster_manager.opensearch_client.is_node_up():
+            self.update_external_client_endpoints(external_client)
+        else:
+            event.defer()
+
+    def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
```
reneradoi, line 161: In this method, I've found errors when trying to update relation data. These errors can also been seen in the integration test run, here's an example: https://github.com/canonical/opensearch-single-kernel-library/actions/runs/23005682132/job/66932424111?pr=18#step:14:11182

reneradoi (commented):
Hey Akram! I've reviewed again, there are a few questions/comments left.

Unrelated to this PR but important: When testing, I found that we leak the CA certificate in the debug logs. This is not the case on the opensearch charm itself, so it was probably introduced in the refactoring work. Please have a look, we must avoid leaking secrets.

Example: https://github.com/canonical/opensearch-single-kernel-library/actions/runs/23005682132/job/66932424111?pr=18#step:14:1542.

opensearch_single_kernel/common/statuses.py
akram09: @Mehdi-Bendriss What do you think ?

opensearch_single_kernel/managers/external_clients.py
akram09: Thank you, nice catch.

opensearch_single_kernel/events/external_clients.py
akram09: Yes, this is how it was implemented previously

opensearch_single_kernel/events/external_clients.py
akram09: It is a safeguard in case we didn't clean it up previously

opensearch_single_kernel/events/external_clients.py
akram09: Yes, I talked with Mehdi about this and we said we will clean it

opensearch_single_kernel/common/client.py
```diff
@@ -48,6 +53,235 @@ def __init__(
...
+            if (
+                e.response_code == 400
+                and e.response_body.get("error", {}).get("type")
+                == "resource_already_exists_exception"
+            ):
+                logger.warning("Index failed to be created as it already exists, continuing...")
+            else:
+                raise e
+
+    def create_role(
```
Mehdi-Bendriss: nit, can we use `_user_role` in all of these methods to differentiate with the node roles?

opensearch_single_kernel/common/constants.py
```diff
@@ -143,12 +191,33 @@ class OpenSearchPaths(BaseStrEnum):
...
+
+# OpenSearch Protected Indices
+PROTECTED_INDEX_NAMES = [
+    ".opendistro_security",
+    ".opendistro-alerting-config",
+    ".opendistro-alerting-alert*",
+    ".opendistro-anomaly-results*",
+    ".opendistro-anomaly-detector*",
+    ".opendistro-anomaly-checkpoints",
+    ".opendistro-anomaly-detection-state",
```
Mehdi-Bendriss, line 217: For another PR, i'm not so sure these are worth protecting - if someone has the admin extra user role, I don't see why they shouldn't see these (exception being on the opendistro-security)

opensearch_single_kernel/common/constants.py
```diff
@@ -210,3 +280,9 @@ class OpenSearchPaths(BaseStrEnum):
 DATA_ROLE_REMOVAL_FORBIDDEN = (
     "Removal of data role from current deployment not allowed - the data cannot be reallocated."
 )
+
+
+# Endpoints
+USER_ENDPOINT = "/_plugins/_security/api/internalusers"
+ROLE_ENDPOINT = "/_plugins/_security/api/roles"
```
Mehdi-Bendriss: nit, similar to a previous comment on "user_" prefix

opensearch_single_kernel/common/statuses.py
Mehdi-Bendriss: Yes, I would probably change the status to something along the lines of `Failed to create {index} index - see the logs...`

opensearch_single_kernel/common/statuses.py
```diff
@@ -69,6 +69,13 @@ class CharmStatuses(Enum):
...
+        "failed to create {index} index - deferring index-requested event..."
+    )
+    USER_CREATION_FAILED = BlockedStatus("failed to create users for {rel_name} relation {id}")
```
Mehdi-Bendriss:
```suggestion
    USER_CREATION_FAILED = BlockedStatus("Failed to create users for {rel_name} relation {id}")
```

opensearch_single_kernel/core/state.py
```diff
@@ -363,6 +368,17 @@ def is_data_role_in_cluster_fleet_apps(self) -> bool:
         data_apps_in_fleet = [app for app in self.apps_in_fleet() if "data" in app.roles]
         return data_apps_in_fleet and any(app.planned_units > 0 for app in data_apps_in_fleet)
 
+    @property
+    def client_users_dict(self) -> dict[str, str]:
+        """Get the client relation users dict from application databag."""
+        users_dict = self.get_object("client_relation_users") or {}
+        return users_dict
```
Mehdi-Bendriss: nit:
```suggestion
        return self.get_object("client_relation_users") or {}
```

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,202 @@
...
+        # TODO: If upgrade in progress then defer event
+
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.cluster_manager.opensearch_client.is_node_up() or not event.index:
+            event.defer()
+            return
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
```
Mehdi-Bendriss: nit, walrus

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,202 @@
...
+            logger.error("No external client found for relation id %d", event.relation.id)
+            return
+
+        if not validate_index_name(event.index):
+            raise OpenSearchIndexError(f"invalid index name: {event.index}")
```
Mehdi-Bendriss: 1. Why not Blocked Status and return instead of crashing the hook? 
2. could you open an issue for this one? in the next cycle, we'll propagate errors from the provider to the requirer instead of displaying BlockedStatuses on the server charm

opensearch_single_kernel/events/opensearch.py
```diff
@@ -169,14 +167,19 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:  # noq
...
-        # self.opensearch_provider.update_endpoints(relation)
-
         self.charm.config_manager.update_seeds_config()
 
         if self.charm.unit.is_leader():
-            nodes = self.charm.cluster_manager.get_nodes(is_node_up)
+            try:
+                nodes = self.charm.cluster_manager.get_nodes(is_node_up)
+            except OpenSearchHttpError:
+                logger.error("unable to get nodes")
```
Mehdi-Bendriss, line 176: also worth logging the full stacktrace

opensearch_single_kernel/events/opensearch.py
```diff
@@ -311,7 +314,8 @@ def _on_update_status(self, event: UpdateStatusEvent) -> None:  # noqa: C901
             without the user noticing in case the cert of the unit transport layer expires.
             So we want to stop opensearch in that case, since it cannot be recovered from.
         """
-        if not self.charm.state.application.deployment_desc:
+        deployment_desc = self.charm.state.application.deployment_desc
+        if not deployment_desc:
```
Mehdi-Bendriss: nit, walrus

opensearch_single_kernel/events/opensearch.py
```diff
@@ -337,26 +341,27 @@ def _on_update_status(self, event: UpdateStatusEvent) -> None:  # noqa: C901
...
-        #    self.unit.is_leader()
-        #    and deployment_desc
-        #    and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
-        # ):
-        #    self.opensearch_provider.remove_lingering_relation_users_and_roles()
+        if self.charm.unit.is_leader():
+            try:
+                nodes = self.charm.cluster_manager.get_nodes(use_localhost=True)
+            except OpenSearchHttpError:
+                logger.error("unable to get nodes")
```
Mehdi-Bendriss: also worth logging the actual error

opensearch_single_kernel/events/opensearch.py
```diff
@@ -492,19 +504,25 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:  # noqa: C901
...
+        try:
+            # Restore purged system users in local `internal_users.yml`
+            # with corresponding credentials
+            if self.charm.unit.is_leader():
+                for user in OPENSEARCH_SYSTEM_USERS:
+                    self.charm.internal_users_manager.put_or_update_internal_user_leader(
+                        user, update=False
+                    )
+        except OpenSearchUserMgmtError:
+            logger.error("An error occurred while updating internal user")
```
Mehdi-Bendriss: any reason not to log the error as well?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1129,6 +1165,19 @@ def request_new_admin_certificate(self) -> None:
...
         self.charm.tls_events.certs.request_certificate_creation(certificate_signing_request=csr)
 
+    def update_external_clients_endpoints(self) -> None:
+        """Update the endpoints of all the external clients relations."""
+        for external_client in self.charm.state.external_clients:
+            if self.charm.unit.is_leader():
+                try:
+                    nodes = self.charm.cluster_manager.get_nodes(use_localhost=True)
+                except OpenSearchHttpError:
+                    logger.error("unable to get nodes")
```
Mehdi-Bendriss: on logging the error as well

opensearch_single_kernel/managers/exclusions.py
```diff
@@ -24,6 +24,8 @@
 class NodesExclusionsManager(BaseManager):
     """OpenSearch Nodes Exclusions Manager."""
 
+    CONFIG_YML = "opensearch.yml"
```
Mehdi-Bendriss: don't we have this somewhere already declared in the opensearch config manager?

opensearch_single_kernel/workload/vm.py
```diff
@@ -122,6 +122,14 @@ def get_host_public_ip(self) -> str | None:
 
         return output.out.strip()
 
+    def is_started(self) -> bool:
+        """Check if OpenSearch is started."""
+        reachable = self.is_reachable(self.host, self.port)
+        if not reachable:
```
Mehdi-Bendriss: nit walrus

Mehdi-Bendriss (commented):
Great work Akram! only small comments and nits 
(pending the client_relations tests to pass)

opensearch_single_kernel/managers/external_clients.py
reneradoi: I think we still need to catch the `OpenSearchCmdError` that this might raise if the called `run_cmd` fails.

reneradoi (commented):
Just one comment left from my side (I've unresolved it conversation about `ExternalClientsManager.version`. Other than that I'm fine with the changes and will approve as soon as the tests are passing.

opensearch_single_kernel/common/client.py
```diff
@@ -48,6 +53,235 @@ def __init__(
...
+        except OpenSearchHttpError as e:
+            if e.response_code == 404:
+                return {
+                    "status": "OK",
+                    "response": "role does not exist, and therefore has not been removed",
+                }
+            else:
+                raise e
+
+        logger.debug(resp)
```
s-mlnk: I think it's left from debugging and should be removed. Same with `remove_user`.

opensearch_single_kernel/common/constants.py
akram09: Ref #42

opensearch_single_kernel/events/external_clients.py
akram09: Ref #43

opensearch_single_kernel/events/opensearch.py
akram09: No, I just missed that.

opensearch_single_kernel/managers/exclusions.py
akram09: Removed

opensearch_single_kernel/workload/vm.py
```diff
@@ -122,6 +122,13 @@ def get_host_public_ip(self) -> str | None:
 
         return output.out.strip()
 
+    def is_started(self) -> bool:
```
s-mlnk: There are no usages of this function and there's no abstract function with this name in base class.

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -2,22 +2,271 @@
...
         super().__init__(state, workload)
         self.name = "external_clients_manager"
-        self.yaml_setter = YamlConfigSetter(self.workload)
+
+    def create_opensearch_users(
+        self,
+        external_client: ExternalOpenSearchClient,
+        index: str,
+        extra_user_roles: str | None = None,
+    ) -> tuple[str, str] | None:
```
s-mlnk: Seems that it always returns value:
```suggestion
    ) -> tuple[str, str]:
```

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            username, pwd = self.charm.external_clients_manager.create_opensearch_users(
+                external_client, event.index, extra_user_roles=event.extra_user_roles
+            )
+        except OpenSearchUserMgmtError as err:
+            logger.error(err)
+            self.charm.status.set(
+                CharmStatuses.USER_CREATION_FAILED,
+                dynamic_params={"rel_name": CLIENT_RELATION, "id": event.relation.id},
+            )
```
s-mlnk, line 112: question: do we need `event.defer()` here just like it is with `create_index` above?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -0,0 +1,208 @@
...
+        external_client = self.charm.state.external_client_by_relation(event.relation)
+        if not external_client:
+            logger.error("No external client found for relation id %d", event.relation.id)
+            return
+        if self.charm.cluster_manager.opensearch_client.is_node_up():
+            self.update_external_client_endpoints(external_client)
+        else:
+            event.defer()
+
+    def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
```
s-mlnk, line 161: question: will the `INDEX_CREATION_FAILED` and `USER_CREATION_FAILED` statuses be cleared in case of removing the relation entirely instead of fixing it?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -446,6 +450,13 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         profile_restart_needed = self.charm.config_manager.update_profile_configuration(
             config_profile
         )
+        if self.charm.unit.is_leader():
+            try:
+                self.charm.external_clients_manager.update_relations_roles_mapping()
+            except OpenSearchUserMgmtError as e:
+                logger.error("Failed to update relations roles mapping: %s", e)
+                event.defer()
```
s-mlnk, line 458: question: missing return here?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -492,19 +503,25 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:  # noqa: C901
...
+            # Restore purged system users in local `internal_users.yml`
+            # with corresponding credentials
+            if self.charm.unit.is_leader():
+                for user in OPENSEARCH_SYSTEM_USERS:
+                    self.charm.internal_users_manager.put_or_update_internal_user_leader(
+                        user, update=False
+                    )
+        except OpenSearchUserMgmtError as e:
+            logger.error("An error occurred while updating internal user %s", str(e))
+            event.defer()
```
s-mlnk, line 525: question: missing return here?

s-mlnk (commented):
Nice work! Just have a couple of questions.

reneradoi (approved):
Very well done!

Mehdi-Bendriss (approved):

PR #23, opened 2026-02-18
patch: [DPE-9185] [DPE-9401] Port plugin & keystore managers and SMTP integration

s-mlnk:
# Key changes

## Plugins

* Ported plugin manager.
* Ported plugin events handler. However, the whole purpose of this events handler is to handle large deployments and while we don't support them, it is non-functional.
* The `get_object` & `put_object` functions were moved from `ApplicationState` to the `RelationState` base class in order to bring that functionality to both application & server contexts. It is used by `plugin_config_info` state property.
* Added plugins integration test with large deployments & COS (#33) tests skipped due to missing support.

## Notifications

* Ported notifications manager & events handler.
* Added SMTP charm library, on which notifications are dependent.
* Added smtp interface in the `tests/charms/opensearch_test_charm/metadata.yaml`.

## Keystore

* Ported keystore manager & events handler.
* Added `opensearch_keystore` path to the workload's `Paths`.

### DIFF
--- a/opensearch_single_kernel/common/client.py
+++ b/opensearch_single_kernel/common/client.py
@@ -602,8 +602,7 @@ class OpenSearchClient:
             self.request("GET", f"/_plugins/_notifications/configs/{config_id}")
             return True
         except OpenSearchHttpError as exc:
-            code = getattr(exc, "response_code", None)
-            if code == 404:
+            if getattr(exc, "response_code", None) == 404:
                 return False
             raise
 
@@ -635,7 +634,8 @@ class OpenSearchClient:
     def delete_notification_config(self, config_id: str) -> None:
         """Delete config by id.
 
-        404 (config already gone) is treated as success.
+        If the request returns code 404 (config already gone)
+        it is treated as success and function returns.
 
         Args:
             config_id: Notification Config ID
@@ -651,7 +651,7 @@ class OpenSearchClient:
         """Reload secure settings. Doesn't throw an exception.
 
         Returns:
-            whether operation was successful.
+            bool: whether operation was successful.
         """
         try:
             response = self.request("POST", "_nodes/reload_secure_settings")
--- a/opensearch_single_kernel/common/client.py
+++ b/opensearch_single_kernel/common/client.py
@@ -33,7 +33,11 @@ class OpenSearchClient:
     """Handle OpenSearch Interaction with Server."""
 
     def __init__(
-        self, workload: BaseWorkload, host: str, port: int, admin_secret: str | None = None
+        self,
+        workload: BaseWorkload,
+        host: str,
+        port: int,
+        admin_secret: str | None = None,
     ):
         """Initialise the client.
 
@@ -554,7 +558,11 @@ class OpenSearchClient:
             raise OpenSearchHttpError(response_text=str(e))
 
     def get_log_error_http_retry(
-        self, retry_max: int, method: str, urls: list[str], payload: dict[str, Any] | None
+        self,
+        retry_max: int,
+        method: str,
+        urls: list[str],
+        payload: dict[str, Any] | None,
     ):
         """Return a custom log function to run before a new Tenacity retry."""
 
@@ -567,3 +575,88 @@ class OpenSearchClient:
             )
 
         return log_error
+
+    def create_notification_config(
+        self, *, config_id: str, name: str, config: dict[str, object]
+    ) -> None:
+        """Create notification config.
+
+        Args:
+            config_id: Notification Config ID
+            name: Notification Name
+            config: Notification Config
+        """
+        payload = {"config_id": config_id, "name": name, "config": config}
+        self.request("POST", "/_plugins/_notifications/configs/", payload=payload)
+
+    def notification_config_exists(self, config_id: str) -> bool:
+        """Check if config exists.
+
+        Args:
+            config_id: Notification Config ID
+
+        Returns:
+            True if config exists, False if 404.
+        """
+        try:
+            self.request("GET", f"/_plugins/_notifications/configs/{config_id}")
+            return True
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code == 404:
+                return False
+            raise
+
+    def create_or_update_notification_config(
+        self, *, config_id: str, name: str, config: dict[str, object]
+    ) -> None:
+        """Create config if missing, otherwise update.
+
+        Args:
+            config_id: Notification Config ID
+            name: Notification Name
+            config: Notification Config
+        """
+        if self.notification_config_exists(config_id):
+            self.update_notification_config(config_id=config_id, config=config)
+        else:
+            self.create_notification_config(config_id=config_id, name=name, config=config)
+
+    def update_notification_config(self, *, config_id: str, config: dict[str, object]) -> None:
+        """Update notification config.
+
+        Args:
+            config_id: Notification Config ID
+            config: Notification Config
+        """
+        payload = {"config": config}
+        self.request("PUT", f"/_plugins/_notifications/configs/{config_id}", payload=payload)
+
+    def delete_notification_config(self, config_id: str) -> None:
+        """Delete config by id.
+
+        404 (config already gone) is treated as success.
+
+        Args:
+            config_id: Notification Config ID
+        """
+        try:
+            self.request("DELETE", f"/_plugins/_notifications/configs/{config_id}")
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code == 404:
+                return
+            raise
+
+    def reload_secure_settings(self) -> bool:
+        """Reload secure settings. Doesn't throw an exception.
+
+        Returns:
+            whether operation was successful.
+        """
+        try:
+            response = self.request("POST", "_nodes/reload_secure_settings")
+        except OpenSearchHttpError as e:
+            logger.error("Could not reload secure settings: %s", e)
+            return False
+        return isinstance(response, dict) and response.get("_nodes", {}).get("failed", -1) == 0
--- a/opensearch_single_kernel/common/client.py
+++ b/opensearch_single_kernel/common/client.py
@@ -607,7 +607,7 @@ class OpenSearchClient:
                 return False
             raise
 
-    def create_or_update_notification_config(
+    def put_notification_config(
         self, *, config_id: str, name: str, config: dict[str, object]
     ) -> None:
         """Create config if missing, otherwise update.
@@ -643,8 +643,7 @@ class OpenSearchClient:
         try:
             self.request("DELETE", f"/_plugins/_notifications/configs/{config_id}")
         except OpenSearchHttpError as exc:
-            code = getattr(exc, "response_code", None)
-            if code == 404:
+            if getattr(exc, "response_code", None) == 404:
                 return
             raise
 
--- a/opensearch_single_kernel/core/models.py
+++ b/opensearch_single_kernel/core/models.py
@@ -21,11 +21,9 @@ from opensearch_single_kernel.common.constants import (
     DeploymentType,
     Directive,
     PerformanceType,
+    SmtpTransportSecurity,
     StartMode,
     State,
-)
-from opensearch_single_kernel.lib.charms.smtp_integrator.v0.smtp import (
-    TransportSecurity,
 )
 
 logger = logging.getLogger(__name__)
@@ -386,9 +384,7 @@ class PluginConfigInfo(Model):
         """Merge items into cleanup dictionary avoiding duplicates."""
         for key, items in cleanup.items():
             current = self.cleanup.setdefault(key, [])
-            for item in items:
-                if item not in current:
-                    current.append(item)
+            current = sorted(list(set(current) | set(items)))
 
 
 @dataclass(frozen=True)
@@ -409,4 +405,4 @@ class SmtpConfig:
     label: str
     group_id: str
     channel_id: str
-    transport_security: TransportSecurity
+    transport_security: SmtpTransportSecurity
--- a/opensearch_single_kernel/core/models.py
+++ b/opensearch_single_kernel/core/models.py
@@ -5,10 +5,10 @@
 
 """Collection of models used for the operation of the charm."""
 
-
 import json
 import logging
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from datetime import datetime
 from hashlib import md5
 from typing import Any, Iterator, Literal
@@ -23,6 +23,9 @@ from opensearch_single_kernel.common.constants import (
     PerformanceType,
     StartMode,
     State,
+)
+from opensearch_single_kernel.lib.charms.smtp_integrator.v0.smtp import (
+    TransportSecurity,
 )
 
 logger = logging.getLogger(__name__)
@@ -308,7 +311,8 @@ class OpenSearchProfile(ABC):
         """Get the JVM heap size in KB based on the memory requirements."""
         if self.memory_requirements.jvm_heap_percentage:
             return min(
-                int(self.memory_requirements.jvm_heap_percentage * mem_size), MAX_HEAP_SIZE_IN_KB
+                int(self.memory_requirements.jvm_heap_percentage * mem_size),
+                MAX_HEAP_SIZE_IN_KB,
             )
         return _1GB_IN_KB
 
@@ -385,3 +389,24 @@ class PluginConfigInfo(Model):
             for item in items:
                 if item not in current:
                     current.append(item)
+
+
+@dataclass(frozen=True)
+class SmtpConfig:
+    """SMTP-related config derived from relation data.
+
+    Attributes:
+        sender_email: From-address for the SMTP sender (relation smtp_sender).
+        smtp_account_id: OpenSearch config id for the SMTP account (e.g. smtp-88_smtp-account).
+        label: Plugin/config label for this relation (e.g. plugin-notifications-88).
+        group_id: OpenSearch config id for the recipient group (e.g. smtp-88_recipients).
+        channel_id: OpenSearch config id for the email channel (e.g. smtp-88_email-channel).
+        transport_security: SMTP transport security (none, start_tls, tls).
+    """
+
+    sender_email: str
+    smtp_account_id: str
+    label: str
+    group_id: str
+    channel_id: str
+    transport_security: TransportSecurity
--- a/opensearch_single_kernel/core/models.py
+++ b/opensearch_single_kernel/core/models.py
@@ -384,7 +384,7 @@ class PluginConfigInfo(Model):
         """Merge items into cleanup dictionary avoiding duplicates."""
         for key, items in cleanup.items():
             current = self.cleanup.setdefault(key, [])
-            current = sorted(list(set(current) | set(items)))
+            self.cleanup[key] = sorted(list(set(current) | set(items)))
 
 
 @dataclass(frozen=True)
--- a/opensearch_single_kernel/core/models.py
+++ b/opensearch_single_kernel/core/models.py
@@ -369,3 +369,19 @@ class TestingProfile(OpenSearchProfile):
             cluster_managers=1,
             data=1,
         )
+
+
+class PluginConfigInfo(Model):
+    """Model class for representing data needed to add or remove plugin configuration"""
+
+    relation_name: str | None = None
+    secret_id: str | None = None
+    cleanup: dict[str, list[str]] = Field(default_factory=dict)
+
+    def add_cleanup_items(self, cleanup: dict[str, list[str]]) -> None:
+        """Merge items into cleanup dictionary avoiding duplicates."""
+        for key, items in cleanup.items():
+            current = self.cleanup.setdefault(key, [])
+            for item in items:
+                if item not in current:
+                    current.append(item)
--- a/opensearch_single_kernel/core/relations.py
+++ b/opensearch_single_kernel/core/relations.py
@@ -233,6 +233,33 @@ class RelationState:
             except KeyError:
                 pass
 
+    def get_object(self, key: str) -> dict[str, Any] | None:
+        """Get dict / json object from the relation data store."""
+        data = self.relation_data.get(key)
+        if data is None:
+            return None
+
+        return json.loads(data)
+
+    def put_object(self, key: str, value: dict[str, Any], merge: bool = False) -> None:
+        """Put dict / json object into relation data store."""
+        if merge:
+            stored = self.get_object(key)
+
+            if stored is not None:
+                stored.update(value)
+                value = stored
+
+        sorted_value = Model.sort_payload(value)
+
+        payload_str = None
+        if value is not None:
+            payload_str = json.dumps(
+                sorted_value, default=RelationDataStore._default_encoder, sort_keys=True
+            )
+
+        self.update({key: payload_str})
+
 
 class PeerClusterOrchestratorData(ProviderData, RequirerData):
     """Orchestrator provider data model."""
--- a/opensearch_single_kernel/core/relations.py
+++ b/opensearch_single_kernel/core/relations.py
@@ -4,6 +4,7 @@
 # See LICENSE file for licensing details.
 
 """Base classes for charm relations."""
+
 import enum
 import json
 import logging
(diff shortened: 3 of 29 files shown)


opensearch_single_kernel/core/relations.py
```diff
@@ -233,6 +233,33 @@ def update(self, items: dict[str, str]) -> None:
             except KeyError:
                 pass
 
+    def get_object(self, key: str) -> dict[str, Any] | None:
+        """Get dict / json object from the relation data store."""
+        data = self.relation_data.get(key)
+        if data is None:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/core/relations.py
```diff
@@ -233,6 +233,33 @@ def update(self, items: dict[str, str]) -> None:
...
+            return None
+
+        return json.loads(data)
+
+    def put_object(self, key: str, value: dict[str, Any], merge: bool = False) -> None:
+        """Put dict / json object into relation data store."""
+        if merge:
+            stored = self.get_object(key)
+
+            if stored is not None:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/core/models.py
```diff
@@ -369,3 +369,19 @@ def cluster_topology_requirements(self) -> ClusterTopologyRequirements:
...
+
+    relation_name: str | None = None
+    secret_id: str | None = None
+    cleanup: dict[str, list[str]] = Field(default_factory=dict)
+
+    def add_cleanup_items(self, cleanup: dict[str, list[str]]) -> None:
+        """Merge items into cleanup dictionary avoiding duplicates."""
+        for key, items in cleanup.items():
+            current = self.cleanup.setdefault(key, [])
+            for item in items:
```
akram09: Why not using set and doing a union ?

opensearch_single_kernel/core/relations.py
```diff
@@ -233,6 +233,33 @@ def update(self, items: dict[str, str]) -> None:
             except KeyError:
                 pass
 
+    def get_object(self, key: str) -> dict[str, Any] | None:
```
akram09, line 237: Let's create an issue when we address secrets to be refactored to remove the `RelationDataStore` and DataStore since we are only using it in secrets management. We should also have a static method here called `_default_encoder` to make sure we are writing correctly data to databag

opensearch_single_kernel/core/secrets.py
```diff
@@ -237,10 +237,30 @@ def delete(self, scope: Scope, key: str) -> None:
 
         logger.debug(f"Deleted secret {scope}:{key}")
 
+    def get_tracked_secret(self, secret_id: str, scope: Scope, key: str) -> Secret | None:
```
akram09, line 240: I don't see why we would need this function ? Seems that `_get_juju_secret` is doing the same thing. The function is also very misleading. We are not really tracking, just refreshing the cached_secrets

opensearch_single_kernel/core/secrets.py
```diff
@@ -237,10 +237,30 @@ def delete(self, scope: Scope, key: str) -> None:
...
+
+        self.cached_secrets.set_meta(scope, label, secret)
+        self.cached_secrets.put_content(scope, label, secret.get_content(refresh=True))
+        return secret
+
     def get_secret_id(self, scope: Scope, key: str) -> str | None:
         """Get the secret ID from the cache."""
         label = self.label(scope, key)
-        return self.charm.peers_data.get(scope, label)
+        return (
```
akram09, line 259: Thank you !

opensearch_single_kernel/core/state.py
```diff
@@ -211,6 +210,23 @@ def last_host_ip(self, value: str) -> None:
         """Set the value of last configured IP for the unit. Used for tracking the IP change."""
         self.update({"last_host_ip": value})
 
+    @property
+    def plugin_config_info(self) -> dict[str, PluginConfigInfo]:
```
akram09, line 214: Would it be possible to have a more structured approach ? Meaning if we have known the keys we would have we could just have a PluginsConfigInfo with the predefined keys

akram09: Now that I think about it, it actually make more sense to me to have it structured. Since we could have a general PluginManager that has base functions to update/read/remove data from the databag. and we could inherit using Specific plugins like NotificationsManager. And in the different hooks of peer_cluster we could use the appropriate manager. It would be less error prone, more readable. 


But I think we are updating the behavior here. I want to have @Mehdi-Bendriss opinion regarding this. I can do the refactor in my Large Deployment PR

opensearch_single_kernel/core/state.py
```diff
@@ -225,33 +241,6 @@ def __init__(
         super().__init__(relation, data_interface, component)
         self.app = component
 
-    def get_object(self, key: str) -> dict[str, Any] | None:
```
akram09, line 228: Great

opensearch_single_kernel/events/keystore.py
```diff
@@ -0,0 +1,41 @@
...
+if TYPE_CHECKING:
+    from opensearch_single_kernel.charms.base import OpenSearchBaseCharm
+
+logger = logging.getLogger(__name__)
+
+
+class KeystoreEventsHandler(Object):
+    """Keystore events."""
+
+    reload_event = EventSource(ReloadKeystoreEvent)
```
akram09: In my Backups PR I have moved this custom event to the charm since we have the other custom events there

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+logger = logging.getLogger(__name__)
+
+
+class NotificationsEvents(Object):
+    """Events handler for smtp events"""
+
+    relation_name = SMTP_RELATION
+
+    def __init__(self, charm: "OpenSearchBaseCharm"):
+        super().__init__(charm, "plugin:notifications")
```
akram09: nit: Follow the same name style of other events

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+        self.framework.observe(self.charm.on.secret_changed, self._on_secret_changed)
+
+    def _on_smtp_credentials_changed(self, event: SmtpDataAvailableEvent) -> None:  # noqa: C901
+        """Configure notifications sender/group/channel and keystore creds for this relation.
+
+        Args:
+            event: Smtp credentials available event
+        """
+        parameters = None
```
akram09: nit: rename to smtp_data

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+        if not parameters.port:
+            missing.append("port")
+        if not parameters.transport_security:
+            missing.append("transport_security")
+        if parameters.auth_type != "none":
+            if not parameters.user:
+                missing.append("user")
+            if not parameters.password:
+                missing.append("password")
+
```
akram09: This can be moved to the manager and throw an exception in case of missing parameters which can be handled here and set status

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,336 @@
...
+from opensearch_single_kernel.common.exceptions import OpenSearchHttpError
+from opensearch_single_kernel.core.state import ClusterState
+from opensearch_single_kernel.lib.charms.smtp_integrator.v0.smtp import SmtpRelationData
+from opensearch_single_kernel.managers.base import BaseManager
+from opensearch_single_kernel.workload.base import BaseWorkload
+
+
+class NotificationsClientError(RuntimeError):
+    """Raises when Notifications API operations fail."""
+
```
akram09: Move to exception folder

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,336 @@
...
+    """
+
+    NONE = "none"
+    STARTTLS = "starttls"
+    TLS = "tls"
+
+    def api_method(self) -> str:
+        """Return the OpenSearch Notifications API method string."""
+        return {"none": "none", "starttls": "start_tls", "tls": "ssl"}[self.value]
+
```
akram09: Move to constants

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,336 @@
...
+        channel_id: OpenSearch config id for the email channel (e.g. smtp-88_email-channel).
+        transport_security: SMTP transport security (none, start_tls, tls).
+    """
+
+    sender_email: str
+    smtp_account_id: str
+    label: str
+    group_id: str
+    channel_id: str
+    transport_security: TransportSecurity
```
akram09: Move to Models

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+            self.charm.status.clear(CharmStatuses.SMTP_CONFIGURATION_ERROR, app=True)
+
+        if parameters.auth_type != "none":
+            # store keystore creds on every unit
+            entries = {
+                f"opensearch.notifications.core.email.{config.smtp_account_id}.username": parameters.user,
+                f"opensearch.notifications.core.email.{config.smtp_account_id}.password": parameters.password,
+            }
+            self.charm.keystore_manager.put_entries(entries)
```
akram09: I had a thought on keystore manager, and I came to a conclusion that it is okey to have the keystore manager be more globally scoped. meaning instead of using it as a utility to put/read/remove entities, it should have context specific functions like `put_notifications_plugin_entries()` since it is a manager and not a utility. I have applied the same idea to the backups PR

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+            if self.charm.unit.is_leader():
+                # leader stores secret for subclusters for per relation
+                self.charm.plugin_manager.store_plugin_secret(
+                    content={
+                        "keys": entries,
+                        "smtp_account_id": cleanup["smtp_account_id"],
+                    },
+                    label=config.label,
+                    relation_name=self.relation_name,
+                )
```
akram09: Can we wrap this in the manager ?

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+        smtp_account_id = smtp_account_ids[0] if smtp_account_ids else None
+
+        # No smtp_account_id; nothing to clean in keystore or notifications
+        if not smtp_account_id:
+            self.charm.plugin_manager.remove_plugin_config(scope=Scope.UNIT, label=label)
+            if self.charm.unit.is_leader():
+                self.charm.plugin_manager.remove_plugin_secret(label)
+                # if self.charm.opensearch_peer_cm.is_provider(typ="main"):
+                #     self.charm.peer_cluster_provider.refresh_relation_data(event)
+            return
```
akram09, line 254: Move to manager

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,345 @@
...
+        if not self.charm.unit.is_leader():
+            return
+
+        self.charm.plugin_manager.remove_plugin_secret(label)
+
+        # if self.charm.opensearch_peer_cm.is_provider(typ="main"):
+        #     self.charm.peer_cluster_provider.refresh_relation_data(event)
+
+    def _on_secret_changed(self, event: SecretChangedEvent) -> None:
+        """Handles secret changes (support multiple smtp relations).
```
akram09: We can explicitely specify which secret we are actually tracking here

opensearch_single_kernel/events/plugin.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Handler for TLS events."""
```
akram09: This is not TLS events

opensearch_single_kernel/events/plugin.py
```diff
@@ -0,0 +1,82 @@
...
+            self.charm.on[PEER_RELATION].relation_changed,
+            self._on_peer_relation_changed,
+        )
+
+    def _on_peer_relation_changed(self, event):  # noqa: C901
+        """Handle plugin secret-related peer relation changes."""
+        # if this is a subcluster, all units must add plugin keys from secrets to their keystores
+        # if not self.charm.opensearch_peer_cm.is_consumer(of="main"):
+        #     return
+        return
```
akram09, line 45: Why are we returning here ?

opensearch_single_kernel/managers/keystore.py
```diff
@@ -0,0 +1,88 @@
...
+            logger.debug("Opensearch not running. Keystore settings will be loaded at start time.")
+            return True
+
+        try:
+            response = self.opensearch_client.request("POST", "_nodes/reload_secure_settings")
+        except OpenSearchHttpError as e:
+            logger.error("Could not reload secure settings: %s", e)
+            return False
+
+        success = response.get("_nodes", {}).get("failed", -1) == 0
```
akram09: This should be at the opensearch client level and we handle the request here. So the except stays here but other things should go to opensearchClient

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        try:
+            self.opensearch_client.request(
+                "DELETE", f"/_plugins/_notifications/configs/{config_id}"
+            )
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code == 404:
+                return
+            raise
```
akram09: Move to openSearchClient, please make sure to follow the same way of doing when it comes to retries (We dont use annotation we use a parameter in the request function)

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,336 @@
...
+        """Update notification config.
+
+        Args:
+            config_id: Notification Config ID
+            config: Notification Config
+        """
+        payload = {"config": config}
+        self.opensearch_client.request(
+            "PUT", f"/_plugins/_notifications/configs/{config_id}", payload=payload
+        )
```
akram09: Same, move to client

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,336 @@
...
+            True if config exists, False if 404.
+        """
+        try:
+            self.opensearch_client.request("GET", f"/_plugins/_notifications/configs/{config_id}")
+            return True
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code == 404:
+                return False
+            raise
```
akram09: Move to client

pyproject.toml
```diff
@@ -120,6 +118,7 @@ allure-pytest-default-results = "^0.1.3"
 # Azure integration tests
 azure-identity = "^1.23.0"
 azure-storage-blob = "^12.25.1"
+oauth_tools = { git = "https://github.com/canonical/iam-bundle", rev = "921d113325eed156e5b54f97c42e6c50810180c0" }
```
akram09: We don't really need this at this point

akram09 (commented):
Thank you Serhii, I left a couple of comments. Please confirm that you have moved the integration tests without any changes from the `opensearch-operator`

opensearch_single_kernel/core/models.py
s-mlnk: If you're talking about changing type in `PluginConfigInfo`: we won't be able to store set in relation as it's not JSON serialisable so it will require a list <> set casting. And if so, it's easier to leave it as-is.
If you're talking about changing this operation to `current = sorted(list(set(current) | set(cleanup)))`, it will be less readable.

opensearch_single_kernel/events/plugin.py
s-mlnk: Because we don't have a proper analogue of the [`is_consumer` check](https://github.com/canonical/opensearch-operator/blob/85bb3aa0e706423288da762bf70358ed16515bd3/lib/charms/opensearch/v0/opensearch_peer_clusters.py#L484) here and I think it's appropriate for the future large deployments PR only.

opensearch_single_kernel/managers/keystore.py
s-mlnk: I moved an entire logic on behalf of returning a result through bool.

opensearch_single_kernel/common/client.py
```diff
@@ -567,3 +575,88 @@ def log_error(retry_state: RetryCallState):
...
+
+        Returns:
+            True if config exists, False if 404.
+        """
+        try:
+            self.request("GET", f"/_plugins/_notifications/configs/{config_id}")
+            return True
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code == 404:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/common/client.py
```diff
@@ -567,3 +575,88 @@ def log_error(retry_state: RetryCallState):
...
+                return False
+            raise
+
+    def create_or_update_notification_config(
```
Mehdi-Bendriss:
```suggestion
    def put_notification_config(
```

opensearch_single_kernel/core/secrets.py
Mehdi-Bendriss: @smethnani do you remember what this was about?

opensearch_single_kernel/core/state.py
Mehdi-Bendriss: I have no objection on the above proposal, I'm okay addressing it in this or in the large deployment PR.

Mehdi-Bendriss (commented):
Thanks Serhii! Great work, I left some minor comments, but the review from Akram is already sound

opensearch_single_kernel/core/models.py
akram09: I think it would be more readable if we just add the proper variables

opensearch_single_kernel/events/plugin.py
akram09: True

opensearch_single_kernel/core/relations.py
akram09: @s-mlnk please don't foreget to create issue for this

opensearch_single_kernel/core/state.py
akram09: Ref. #36

opensearch_single_kernel/common/client.py
s-mlnk: Great catch!

opensearch_single_kernel/events/notifications.py
s-mlnk: 1. There are couple of returns here which we have to signalise event handler with.
2. There's label -> plugin_config -> cleanup -> keys dependency that is used later for keystore manager and therefore cannot be wrapped only in the manager.
3. We also would need to signalise event handler with large deployments part here.

All above will complicate potential wrapped plugin function (especially return params) so much that it makes no sense to move it from here.

opensearch_single_kernel/events/notifications.py
s-mlnk: Changed description to `Handles smtp secret changes (support multiple smtp relations).`.

opensearch_single_kernel/common/client.py
akram09: This don't seem to be resolved

opensearch_single_kernel/common/client.py
```diff
@@ -567,3 +575,87 @@ def log_error(retry_state: RetryCallState):
...
+            config_id: Notification Config ID
+            config: Notification Config
+        """
+        payload = {"config": config}
+        self.request("PUT", f"/_plugins/_notifications/configs/{config_id}", payload=payload)
+
+    def delete_notification_config(self, config_id: str) -> None:
+        """Delete config by id.
+
+        404 (config already gone) is treated as success.
```
akram09: nit: Update this comments to give more context.  ex "If the request returns 404 meaning config already gone we return since this is treated as success"

opensearch_single_kernel/common/client.py
```diff
@@ -567,3 +575,87 @@ def log_error(retry_state: RetryCallState):
...
+        except OpenSearchHttpError as exc:
+            if getattr(exc, "response_code", None) == 404:
+                return
+            raise
+
+    def reload_secure_settings(self) -> bool:
+        """Reload secure settings. Doesn't throw an exception.
+
+        Returns:
+            whether operation was successful.
```
akram09: nit: 
bool: whether operation was successful

opensearch_single_kernel/core/state.py
```diff
@@ -211,6 +210,23 @@ def last_host_ip(self, value: str) -> None:
         """Set the value of last configured IP for the unit. Used for tracking the IP change."""
         self.update({"last_host_ip": value})
 
+    @property
+    def plugin_config_info(self) -> dict[str, PluginConfigInfo]:
+        """Returns configuration information for plugins this app is managing"""
```
akram09: nit: This is a server property not app level

opensearch_single_kernel/events/notifications.py
```diff
@@ -0,0 +1,317 @@
...
+                return
+
+        if self.charm.unit.is_leader():
+            self.charm.status.clear(
+                CharmStatuses.SMTP_COULD_NOT_READ_DATA,
+                app=True,
+                pattern=Status.CheckPattern.Interpolated,
+            )
+
+        if not smtp_data:
```
akram09, line 109: nit: if not smtp_data and is_leader()
Also create a `is_leader` property at the top of function and use it in the function

opensearch_single_kernel/managers/keystore.py
```diff
@@ -0,0 +1,102 @@
...
+        """Add new key/val entries on the keystore."""
+        for key, val in entries.items():
+            # adding the '--force' flag will create the keystore if not present
+            self.workload.run_cmd(self.KEYSTORE, f"add {key} --force", stdin=val)
+
+    def put_file_entry(self, key: str, filename: str) -> None:
+        """Add a new file entry in the keystore."""
+        self.workload.run_cmd(self.KEYSTORE, f"add-file {key} {filename} --force")
+
+    def put_notifications_plugin_smtp_credentials(
```
akram09, line 48: Thanks !

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,233 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""OpenSearch Notifications plugin API client (configs CRUD).
+
+This client wraps OpenSearchDistribution.request() to manage notifications configs:
+- smtp sender (config_type: smtp_account)
+- email group (config_type: email_group)
+- email channel (config_type: email)
```
akram09: Update the description

opensearch_single_kernel/managers/notification.py
```diff
@@ -0,0 +1,233 @@
...
+)
+from opensearch_single_kernel.core.models import SmtpConfig
+from opensearch_single_kernel.core.state import ClusterState
+from opensearch_single_kernel.lib.charms.smtp_integrator.v0.smtp import SmtpRelationData
+from opensearch_single_kernel.managers.base import BaseManager
+from opensearch_single_kernel.workload.base import BaseWorkload
+
+
+class NotificationsManager(BaseManager):
+    """Notifications plugin API client using OpenSearchDistribution request."""
```
akram09: Same

akram09 (dismissed):
Thank you Serhii, only nit comments for me

opensearch_single_kernel/common/client.py
s-mlnk: Sorry, missed that one :(

opensearch_single_kernel/events/notifications.py
s-mlnk: While I addressed `is_leader` suggestion, I cannot combine this two checks because there is a return statement that is under the first check, but after the second one:

```python
        if not smtp_data:
            if is_leader:
                self.charm.status.set(CharmStatuses.SMTP_NO_RELATION_DATA, app=True)
            return # <--
```

opensearch_single_kernel/core/secrets.py
smethnani: It was initially implemented to add a local label to track the secret but we realized if the secret was deleted then the label can't be updated. I believe this can indeed be removed now, thanks @akram09!

s-mlnk: All of the our secret functions are oriented on the secret labels, but in this case we need to retrieve secret by id. Should I create a separate function (alternative to the `get_tracked_secret`) or modify existing ones (e.g. `key_type: Literal["label", "id"]`)?

opensearch_single_kernel/events/notifications.py
akram09: You are right !

opensearch_single_kernel/core/secrets.py
akram09: I would recommend then to take this as a separate issue to be addressed in the secrets Management PR you will be working on @s-mlnk

s-mlnk: #38

opensearch_single_kernel/core/relations.py
s-mlnk: #39

akram09 (approved):

gatici (approved):

Mehdi-Bendriss (approved):

PR #24, opened 2026-02-19
patch: [DPE-8939] Refactor config manager

s-mlnk:
Superseedes #17.

# Key changes

## `opensearch.yml` config refactor

Currently, there are many functions for updating & cleaning specific parts of the `opensearch.yml` config all across the codebase. It results in spaghetti-like config manipulation that is hard to keep track of. This PR refactors it into the single "generate config" function that will improve code readability, will be less prone to development errors and will simplify further development.

## `unicast_hosts.txt` aka `seed hosts` config refactor

Currently, seed hosts are set both from state's `nodes_config` and directly from API in chaotic, unorganised way. This PR refactors it to primarily use `nodes_config` source. The only exception is when `update_opensearch_config` is fed with overridden bootstrap-specific `roles` & `cm_names` & `cm_ips` values.

## Static config options refactor

Currently, static config options are set both along with "dynamic" values (e.g. `ConfigManager.set_node` updates both dynamic and static options) and from distinct "set static options" functions (e.g. `ConfigManager.set_client_auth`). This PR moves all of the "set static options" operations into separate functions and moves their calls to the single place - `ConfigManager.update_opensearch_config`. This removes event handlers responsibility of updating those and improves codebase readability.


opensearch_single_kernel/events/opensearch.py
```diff
@@ -163,12 +164,10 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
         # for relation in self.model.relations.get(ClientRelationName, []):
         # self.opensearch_provider.update_endpoints(relation)
 
-        # register new cm addresses on every node
-        self.charm.config_manager.add_cm_addresses_to_conf()
-
         if self.charm.unit.is_leader():
             # Recompute the node roles in case self-healing didn't trigger leader related event
             self.charm.cluster_manager.recompute_roles_if_needed()
+            self.charm.config_manager.update_nodes_config()
```
akram09: Why we are updating nodes config on leader only ?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
+        """Load the opensearch.yml config of the node."""
+        return self.yaml_setter.load(self.CONFIG_YML)
 
-    def set_admin_tls_conf(self, secrets: dict[str, Any]) -> None:
-        """Configures the admin certificate."""
-        self.yaml_setter.put(
-            self.CONFIG_YML,
-            "plugins.security.authcz.admin_dn/{}",
-            normalized_tls_subject(secrets["subject"]),
+    def _update_static_jvm_options(self):
```
akram09: Why are we using the yaml setter here ? can't this be included in the reconcile function ?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-        if "cluster_manager" in roles and contribute_to_bootstrap:  # cluster NOT bootstrapped yet
-            self.yaml_setter.put(
-                self.CONFIG_YML, "cluster.initial_cluster_manager_nodes", cm_names
-            )
+        self.state.server.last_host_ip = self.state.host_ip
 
-        self.yaml_setter.put(self.CONFIG_YML, "path.data", str(self.workload.paths.data))
-        self.yaml_setter.put(self.CONFIG_YML, "path.logs", str(self.workload.paths.logs))
+        if (
+            self.workload.paths.opensearch_config.exists()
```
akram09: nit question for @Mehdi-Bendriss : Do we need the comments ? Since with this approach we are overriding the comments.But I don't think we have the same problem of ruaml throwing exceptions.

akram09 (commented):
Adding back my unresolved comments to this PR

opensearch_single_kernel/managers/config.py
s-mlnk: > Reconcile function done only for the `opensearch.yml` file. I propose to refactor security config (into reconcile method) later along with JWT or OIDC PR.

s-mlnk: (original comment was about `_update_static_security_options`)

opensearch_single_kernel/managers/config.py
Mehdi-Bendriss: Yeah, we should keep the comments, as they are helpful during debugging/testing sessions. We can replace the `read_text` with an equivalent that loads the whole json. So we compare dicts, but do not override the doc structure

opensearch_single_kernel/events/opensearch.py
s-mlnk: Fixed

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-    def __init__(self, state: ClusterState, workload: BaseWorkload):
-        super().__init__(state, workload)
-        self.name = "config_manager"
+    def __init__(
+        self,
+        state: ClusterState,
+        workload: BaseWorkload,
+    ):
+        self.state = state
+        self.workload = workload
```
gatici: Here, we are inheriting from BaseManager bu stopped calling super().__init__(state, workload). However, in the the file we still use opensearch_client and alt_hosts which are property of BaseManager in the current_node(). We need to call super().__init__(state, workload).

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-            self.yaml_setter.put(
-                self.CONFIG_YML, "cluster.initial_cluster_manager_nodes", cm_names
-            )
+        self.state.server.last_host_ip = self.state.host_ip
 
-        self.yaml_setter.put(self.CONFIG_YML, "path.data", str(self.workload.paths.data))
-        self.yaml_setter.put(self.CONFIG_YML, "path.logs", str(self.workload.paths.logs))
+        if (
+            self.workload.paths.opensearch_config.exists()
+            and self.workload.paths.opensearch_config.read_text() == content
```
gatici: We are comparing read_text() with the content produces with yaml.dump().  yaml.dump does not protect the order, formatting. Even the order changes, comparison will fail and it will detected like a new update which will cause unnecessary restarts.
To prevent this we can use yaml.safe_dump with keys are sorted:
```
yaml.safe_dump(data, sort_keys=True)
```

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-            self.SECURITY_CONFIG_YML,
-            "config/dynamic/authc/basic_internal_auth_domain/http_enabled",
-            True,
+    def _opensearch_tls_config(self, cert_type: CertType) -> dict[str, Any]:
+        layer = "http" if cert_type == CertType.UNIT_HTTP else "transport"
+
+        return (
+            {
+                f"plugins.security.ssl.{layer}.keystore_type": "PKCS12",
+                f"plugins.security.ssl.{layer}.keystore_filepath": f"{self.workload.paths.certs_relative}/{cert_type}.p12",
```
gatici:
```suggestion
                f"plugins.security.ssl.{layer}.keystore_filepath": f"{self.workload.paths.certs_relative}/{cert_type.val}.p12",
```

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
 
-    def set_admin_tls_conf(self, secrets: dict[str, Any]) -> None:
-        """Configures the admin certificate."""
-        self.yaml_setter.put(
-            self.CONFIG_YML,
-            "plugins.security.authcz.admin_dn/{}",
-            normalized_tls_subject(secrets["subject"]),
+    def _update_static_jvm_options(self):
+        self.yaml_setter.replace(self.JVM_OPTIONS, "=logs/", f"={self.workload.paths.logs}/")
+        self.yaml_setter.append(
```
gatici, line 240: This will append the same line again and again. We should check before append to make it idempotent.

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-            self.CONFIG_YML,
-            f"plugins.security.ssl.{target_conf_layer}.enabled_protocols",
-            "TLSv1.2",
-        )
+    def _update_seed_hosts(self, cm_ips: list[str] | None) -> None:
+        """Add CM nodes ips / host names to the seed host list of this unit."""
+        # only update the file if there is data to update
+        if not cm_ips:
+            return
+        cm_ips_set = set(cm_ips)
```
gatici, line 279: The order is not guaranteed with the set. We need to sort it in the next line.

gatici (commented):
Good work @s-mlnk . Now,  state becomes our single source of truth and creating configuration become easier.
I added some comments.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -422,7 +422,11 @@ def _on_install(self, event: InstallEvent) -> None:
 
     def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """On config changed event. Useful for IP changes or for user provided config changes."""
-        if self.charm.config_manager.update_host_if_needed():
+        if (
+            self.charm.state.server.last_host_ip
+            and self.charm.state.host_ip != self.charm.state.server.last_host_ip
```
Mehdi-Bendriss, line 406: could you explain the difference between using the `state` object directly versus using it from the `state.server` with regards to the IPs?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -708,11 +705,13 @@ def _on_start_opensearch(self, event: StartOpenSearch) -> None:  # noqa: C901
         ):
             logger.info("Conditions not met to start opensearch. Will retry next event.")
             event.defer()
+            self.charm.lock_manager.release()
             return
 
         if not self.unit_allowed_to_start(event):
             logger.info("The unit is not allowed to wait, the event need to be retried later.")
             event.defer()
+            self.charm.lock_manager.release()
```
Mehdi-Bendriss: good catch!

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
+
+        Args:
+            roles: override roles got from nodes_config.
+            cm_names: cluster manager nodes for bootstrapping.
+            cm_ips: override seed_hosts got from nodes_config.
+
+        Returns:
+            whether the opensearch.yml config was changed.
+        """
+        if roles is None:
```
Mehdi-Bendriss, line 57: to account for an empty list (in opensearch, an empty roles entry maps to `coordinating`-only)
```suggestion
        if not roles:
```

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
+        return True
 
-        self.yaml_setter.put(self.CONFIG_YML, "plugins.security.disabled", False)
-        self.yaml_setter.put(self.CONFIG_YML, "plugins.security.ssl.http.enabled", True)
-        self.yaml_setter.put(
-            self.CONFIG_YML,
-            "plugins.security.ssl.transport.enforce_hostname_verification",
-            True,
+    @staticmethod
+    def _opensearch_static_config() -> dict[str, Any]:
```
Mehdi-Bendriss, line 85: Missing doctsrings in all methods of this manager

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-        )
+    def _opensearch_host_config(self) -> dict[str, Any]:
+        return {"network.publish_host": self.state.host_ip} if self.state.host_ip else {}
 
-        # enable hot reload of TLS certs (without restarting the node)
-        self.yaml_setter.put(
-            self.CONFIG_YML,
-            "plugins.security.ssl_cert_reload_enabled",
-            True,
+    def _openseearch_temperature_config(self) -> dict[str, Any]:
```
Mehdi-Bendriss:
```suggestion
    def _opensearch_temperature_config(self) -> dict[str, Any]:
```

opensearch_single_kernel/core/state.py
```diff
@@ -195,6 +195,15 @@ def delete_voting_exclusions(self, value: set[str]):
         """Set the value of 'delete_voting_exclusions' in application databag."""
         self.update({"delete-voting-exclusions": ",".join(value)})
 
+    @property
+    def last_host_ip(self) -> str | None:
+        """Get the last configured IP for the unit. Used for checking the IP change."""
+        return self.relation_data.get("last_host_ip")
+
+    @last_host_ip.setter
+    def last_host_ip(self, value: str) -> None:
```
Mehdi-Bendriss, line 210: missing docstrings

opensearch_single_kernel/events/opensearch.py
```diff
@@ -169,8 +170,7 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
         # for relation in self.model.relations.get(ClientRelationName, []):
         # self.opensearch_provider.update_endpoints(relation)
 
-        # register new cm addresses on every node
-        self.charm.config_manager.add_cm_addresses_to_conf()
+        self.charm.config_manager.update_nodes_config()
```
Mehdi-Bendriss: so the `seed_hosts.txt` is managed though this method as well?

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss, line 406: How will this ever happen, when the new ip hasn't been stored yet in the state? the first hook to be invoked when an IP change happens in a unit is this hook

opensearch_single_kernel/events/opensearch.py
```diff
@@ -460,8 +464,8 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
             event.defer()
             return
 
-        profile_restart_needed = self.charm.config_manager.set_profile_configuration_if_needed(
-            current_profile, config_profile
+        profile_restart_needed = self.charm.config_manager.update_profile_configuration(
+            config_profile
```
Mehdi-Bendriss, line 447: good catch

opensearch_single_kernel/events/opensearch.py
```diff
@@ -598,14 +603,6 @@ def _on_start(self, event: StartEvent) -> None:  # noqa: C901
             for user in OPENSEARCH_SYSTEM_USERS:
                 self.charm.users_manager.save_user_locally(user)
 
-        # Configure Client Authentication
```
Mehdi-Bendriss, line 580: why is this no longer needed?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
+        return (
+            {"node.attr.temp": self._opensearch_data_temperature}
+            if self._opensearch_data_temperature
+            else {}
         )
 
-    def cleanup_initial_cluster_managers(self) -> None:
-        """Update the opensearch.yaml by deleting initiali_cluster_manager_nodes."""
-        self.yaml_setter.delete(self.CONFIG_YML, "cluster.initial_cluster_manager_nodes")
+    def _opensearch_cluster_manager_config(
```
Mehdi-Bendriss, line 155: Could we hint that this only during the bootstrapping phase of the cluster?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
+            if (deployment_desc := self.state.application.deployment_desc)
+            else None
         )
 
-        self.yaml_setter.put(
-            self.SECURITY_CONFIG_YML,
-            "config/dynamic/authc/clientcert_auth_domain/transport_enabled",
-            True,
+    @property
+    def _opensearch_tls_subject(self) -> str | None:
```
Mehdi-Bendriss: question: should this (and the following property not be in a `state.tls` object?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
+                    Scope.APP, CertType.APP_ADMIN.val, peek=True
+                )
+            )
+            and (truststore_pwd := admin_secrets.get("truststore-password"))
+            else None
         )
 
-    def update_host_if_needed(self) -> bool:
-        """Update the opensearch config with the current network hosts, after having started.
+    def _opensearch_keystore_pwd(self, cert_type: CertType) -> str | None:
```
Mehdi-Bendriss: could we add the word `tls` in this method? to avoid the confusion between the "opensearch keystore" and the "tls keystore"

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
         )
 
-        if not (new_node_conf := nodes_config.get(self.state.unit_name)):
-            # the conf could not be computed / broadcast, because this node is
-            # "starting" and is not online "yet" - either barely being configured (i.e. TLS)
-            # or waiting to start.
-            return False
+    def _opensearch_roles(self) -> list[str]:
+        if node := self._opensearch_node_config():
+            return node.roles or ["coordinating"]
```
Mehdi-Bendriss: This is not valid (it was a bug before). The `coordinating` role doesn't exist in the **config** of OpenSearch, the way to set a coordinating role is to let the list be empty:
`node.roles: []`

Though in the state we could have `coordinating` to differentiate between "roles that are unset => generated or not set _yet_" VS "coordinating"

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-            # no conf change (roles for now)
-            return False
-        return True
+        if self.state.application.deployment_desc:
+            return self.state.computed_roles()
 
-    def add_seed_hosts(self, cm_ips: list[str]) -> None:
-        """Add CM nodes ips / host names to the seed host list of this unit."""
-        cm_ips_set = set(cm_ips)
+        return ["coordinating"]
```
Mehdi-Bendriss: similar comment here

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
 
-        # only update the file if there is data to update
-        if cm_ips_set:
-            lines = "\n".join([entry for entry in cm_ips_set if entry.strip()])
-            self.workload.write_text(f"{lines}\n", self.workload.paths.seed_hosts)
+    def load_node(self):
```
Mehdi-Bendriss: missing return type hint

opensearch_single_kernel/managers/config.py
```diff
@@ -37,272 +40,289 @@ class ConfigManager(BaseManager):
...
-    def __init__(self, state: ClusterState, workload: BaseWorkload):
-        super().__init__(state, workload)
-        self.name = "config_manager"
+    def __init__(
+        self,
+        state: ClusterState,
+        workload: BaseWorkload,
+    ):
+        self.state = state
+        self.workload = workload
```
Mehdi-Bendriss: I think this should be replaced with:

```suggestion
        super().__init__(state, workload)
        self.name = "config_manager"
```

Mehdi-Bendriss (commented):
Thanks Serhii! Good work! I have left some comments (also pending integration tests to pass)

opensearch_single_kernel/managers/config.py
s-mlnk: Done, but will wait for base PR stability to test it practically.

opensearch_single_kernel/managers/config.py
s-mlnk: It is used to check if roles value was overridden by using function parameter. So it more convenient to be like "did user override roles" and not like "did user override roles with non-empty list".

opensearch_single_kernel/events/opensearch.py
s-mlnk: `update_nodes_config()` manages only `seed_hosts.txt` config file by passing `self.state.application.nodes_config` to `_update_seed_hosts()` function.

opensearch_single_kernel/managers/config.py
s-mlnk: sort_keys is True by-default in yaml.dump.

opensearch_single_kernel/managers/config.py
s-mlnk: Considering this is the only usage of `YamlSetter.append` function, I updated it to check if line already exists prior to writing.

opensearch_single_kernel/events/opensearch.py
s-mlnk: `state.host_ip` returns actual IP from `self.model.get_binding(PEER_RELATION).network.bind_address`, while `state.server.last_host_ip` returns the last IP address written to the config. Latter is used only for the ip changed check.

s-mlnk: https://github.com/canonical/opensearch-single-kernel-library/pull/17#discussion_r2811501927

It should be performed only when IP is changed and not when it set at the first time. And this value is set in ConfigManager during config update.

opensearch_single_kernel/events/opensearch.py
s-mlnk: Because it was setting only static options, that are set now inside `_opensearch_static_config` & `_update_static_security_options` on every `update_opensearch_config` run.

opensearch_single_kernel/managers/config.py
s-mlnk: Are you talking about comments in the `opensearch.yml` file?

opensearch_single_kernel/managers/config.py
s-mlnk: I moved both of them to the state.

opensearch_single_kernel/managers/config.py
s-mlnk: I've added a comment about this in function's docstring.

opensearch_single_kernel/managers/config.py
akram09: yes @s-mlnk

s-mlnk:
~~WIP~~ Done:
- modifying config file with ruaml instead of writing a newly generated one;
- removing the lock_manager release operations added in this PR and avoiding softlock in another way.

opensearch_single_kernel/managers/config.py
s-mlnk: UPD: was changed to ruamel YAML

opensearch_single_kernel/core/state.py
```diff
@@ -623,6 +639,43 @@ def current_peer_cluster_app(self) -> PeerClusterApp:
             ),
         )
 
+    @property
+    def tls_truststore_password(self) -> str | None:
```
akram09, line 658: If we are going to get the passwords from the state. I would personally recommend to have a `TlsState` object defined in the models and added to state where we group these properties and the ones related to TLS

opensearch_single_kernel/core/state.py
```diff
@@ -623,6 +639,43 @@ def current_peer_cluster_app(self) -> PeerClusterApp:
...
+        """Get the keystore-password of provided cert_type from the TLS cert_secret."""
+        return (
+            keystore_pwd
+            if (cert_secret := self.secrets.get_object(Scope.UNIT, cert_type.val, peek=True))
+            and (keystore_pwd := cert_secret.get("keystore-password"))
+            else None
+        )
+
+    @property
+    def tls_subject(self) -> str | None:
```
akram09, line 681: This did confuse me a bit because I remember we have a function in the TLSManager to get subject based on the certificate type,  that function though creates the subject (in a sense) and then we set it to the admin_secrets and then we can read it from there. I would recommend to rename one of the functions to avoid confusion. 
And I would love in a separate PR (or in the k8s PR) to organize this using models

opensearch_single_kernel/events/opensearch.py
```diff
@@ -148,6 +148,10 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:  # noq
             logger.debug("Deployment description not yet computed.")
             return
 
+        if not self.charm.state.server.started:
```
akram09, line 151: Let's not rely only on the started flag but also on `is_node_up()` just to be safe, also why not defer ?

opensearch_single_kernel/managers/config.py
```diff
@@ -37,273 +33,278 @@ class ConfigManager(BaseManager):
...
-            and current_conf.get("node.attr.temp") == new_node_conf.temperature
-        ):
-            # no conf change (roles for now)
-            return False
-        return True
-
-    def add_seed_hosts(self, cm_ips: list[str]) -> None:
-        """Add CM nodes ips / host names to the seed host list of this unit."""
-        cm_ips_set = set(cm_ips)
+    def _opensearch_roles(self) -> list[str]:
```
akram09, line 223: Add @property

opensearch_single_kernel/managers/config.py
```diff
@@ -37,273 +33,278 @@ class ConfigManager(BaseManager):
...
+        return (
+            deployment_desc.config.data_temperature
+            if (deployment_desc := self.state.application.deployment_desc)
+            else None
+        )
 
-        # update (append) CM IPs
-        self.add_seed_hosts(
-            [node.ip for node in list(nodes_config.values()) if node.is_cm_eligible()]
+    def _opensearch_node_config(self) -> Node | None:
```
akram09: Add @property

opensearch_single_kernel/managers/config.py
```diff
@@ -37,273 +33,278 @@ class ConfigManager(BaseManager):
...
 
-    def set_profile_configuration_if_needed(
-        self, current_profile: OpenSearchProfile, config_profile: OpenSearchProfile
-    ) -> bool:
-        """Configure the profile and return whether restart is needed or not"""
-        logger.debug("current profile: %s, config profile: %s", current_profile, config_profile)
-        if current_profile is None or current_profile != config_profile:
-            self.set_jvm_heap_size(
-                config_profile.get_jvm_heap_size(self.workload.meminfo()["MemTotal"])
+    def update_nodes_config(self) -> None:
```
akram09: We are not really updating nodes_config here

opensearch_single_kernel/utils/config.py
```diff
@@ -218,6 +232,25 @@ def put(
...
+        """Rewrite a config file with new configuration while preserving formatting and comments.
+
+        Args:
+            config_file: path to the targeted config while relative to the base path.
+            val: new configuration dictionary.
+
+        Returns:
+            whether config file was changed.
+        """
+        target = self.load(config_file)
```
akram09, line 246: Does this work fine if the file is empty ?

akram09 (commented):
Hello Serhii, 
Thank you for the work, Seems good for me. I have few minor comments

opensearch_single_kernel/utils/config.py
s-mlnk: There's a workaround already in self.load on this case:
```python
        # We are adding a random key:value to the end of the yaml
        # To make sure that the yaml reader will be able to read the file
        # without throwing an exception since sometimes the file might be empty
        # or filled only with comments.
        random_id = uuid.uuid4().hex
        lines.append(f"{random_id}: {random_id}")

        data = self.yaml.load(StringIO("\n".join(lines)))
        del data[random_id]

```

opensearch_single_kernel/managers/config.py
s-mlnk: I've renamed it to the `update_seeds_config`.

opensearch_single_kernel/events/opensearch.py
s-mlnk: I renamed it to `update_seeds_config` to avoid confusion.

opensearch_single_kernel/managers/config.py
s-mlnk: Moved it to the state as property.

opensearch_single_kernel/core/state.py
s-mlnk: Let's do this in different PR as it will add-up irrelevant changes here

opensearch_single_kernel/events/opensearch.py
s-mlnk: There's check below on `is_node_up()`. I've added defer.

opensearch_single_kernel/core/state.py
```diff
@@ -455,6 +471,16 @@ def unit_name(self):
         """Name of the current unit."""
         return format_unit_name(self.model.unit, app=self.application.deployment_desc.app)
 
+    @property
+    def node_config(self) -> Node | None:
```
Mehdi-Bendriss, line 480: is this needed? I remember us having a `current_node` property somewhere

opensearch_single_kernel/core/state.py
```diff
@@ -272,12 +288,12 @@ def is_security_index_initialised(self, value: bool):
         self.update({"security_index_initialised": str(value)})
 
     @property
-    def nodes_config(self) -> dict:
+    def nodes_config(self) -> dict[str, Node]:
```
Mehdi-Bendriss, line 291: nice

opensearch_single_kernel/managers/config.py
```diff
@@ -37,273 +33,270 @@ class ConfigManager(BaseManager):
...
+            roles: override node roles got from nodes_config.
+            cm_names: cluster manager nodes for bootstrapping.
+            cm_ips: override seed_hosts got from nodes_config.
+
+        Returns:
+            whether the opensearch.yml config was changed.
+        """
+        if roles is None:
+            roles = self._opensearch_roles
+
```
Mehdi-Bendriss, line 59: here we need to add a check:
```
if roles == ["coordinating"]:
    roles = []
```
Because internally in the charm we use the role "coordinating" (as a string), but in OpenSearch.yml to set the coordinating role, we must write an empty list. This is an edge case.

opensearch_single_kernel/managers/config.py
Mehdi-Bendriss: This comment still applies

Mehdi-Bendriss (commented):
Thank you Serhii! I left a few small comments, but good to go for me once they're resolved

opensearch_single_kernel/managers/config.py
s-mlnk: But I've already removed initialiser here

opensearch_single_kernel/core/state.py
s-mlnk: It fits into [this discussion in base PR](https://github.com/canonical/opensearch-single-kernel-library/pull/16#discussion_r2863800111).

opensearch_single_kernel/managers/config.py
s-mlnk: It is already done here:
https://github.com/canonical/opensearch-single-kernel-library/blob/ccbcefb290ca631187e400adddfcc87406db760c/opensearch_single_kernel/core/state.py#L647-L648

opensearch_single_kernel/core/state.py
akram09: Alright, please create an issue for that

opensearch_single_kernel/utils/config.py
akram09: Awesome!

akram09 (dismissed):
Seems good for me, Thank you @s-mlnk

opensearch_single_kernel/core/state.py
s-mlnk: Moved `ConfigManager.current_node` to `ExclusionsManager.api_or_state_node`.

opensearch_single_kernel/managers/config.py
Mehdi-Bendriss: I mean that the name of the name of the current manager is missing `self.name = ...`

opensearch_single_kernel/core/state.py
s-mlnk: #32

opensearch_single_kernel/core/state.py
s-mlnk: #32

Mehdi-Bendriss (approved):

akram09 (approved):

gatici (approved):

gatici:
Thank you @s-mlnk !

PR #29, opened 2026-02-26
patch: [DPE-9177] [DPE-9176] Port JWT & OAuth

s-mlnk:
This PR depends on #24.

# Key changes

## JWT

* Ported JWT events handler.
* The `get_object` & `put_object` functions were moved from `ApplicationState` to the `RelationState` base class in order to bring that functionality to both application & server contexts. It is used by `jwt_auth_configuration` state property.

## OAuth

* Ported OAuth events handler.
* Added OAuth charm library, on which events handler is dependent.

## Config manager

* Refactored config manager to reconcile `opensearch-security/config.yml` config instead of updating it in different places, that will improve codebase readability. This actually completes a full config manager refactor.
* All of the default security plugins are preserved in opensearch security to avoid any functional difference between OG and single-kernel Opensearch charms.

## Paths

* Moved `Paths.certs / "chain.pem"` path to `Paths.certs_chain` property.
* Modified TLS & config manager to use new path property.
* Modified Opensearch client to use new path property.

## Cluster manager

Ported `update_security_config` function for JWT & OAuth.

### DIFF
--- a/opensearch_single_kernel/charmcraft.yaml
+++ b/opensearch_single_kernel/charmcraft.yaml
@@ -18,3 +18,5 @@ bases:
 charm-libs:
   - lib: grafana_agent.cos_agent
     version: "0"
+  - lib: hydra.oauth
+    version: "0"
--- a/opensearch_single_kernel/core/relations.py
+++ b/opensearch_single_kernel/core/relations.py
@@ -278,3 +278,14 @@ class PeerCluster(RelationState):
 
     def __init__(self, relation, data_interface, component):
         super().__init__(relation, data_interface, component)
+
+
+class JwtData(RequirerData):
+    """Data Interface to JWT relation on requirer side."""
+
+    def __init__(self, model, relation_name: str):
+        super().__init__(
+            model,
+            relation_name,
+            additional_secret_fields=["signing-key"],
+        )
--- a/opensearch_single_kernel/core/state.py
+++ b/opensearch_single_kernel/core/state.py
@@ -16,8 +16,10 @@ from ops import Application, JujuVersion, Object, Relation, Unit
 from opensearch_single_kernel.common.constants import (
     CLIENT_RELATION,
     GENERATED_ROLES,
+    JWT_CONFIG_RELATION,
     KIBANA_SERVER_ROLE,
     NODE_LOCK_RELATION,
+    OAUTH_RELATION,
     OPENSEARCH_HTTP_PORT,
     PEER_CLUSTER_ORCHESTRATOR_RELATION,
     PEER_CLUSTER_RELATION,
@@ -32,6 +34,7 @@ from opensearch_single_kernel.common.constants import (
 from opensearch_single_kernel.core.models import (
     DeploymentDescription,
     DeploymentType,
+    JWTAuthConfiguration,
     Node,
     OpenSearchProfile,
     PeerClusterApp,
@@ -41,6 +44,7 @@ from opensearch_single_kernel.core.models import (
     TestingProfile,
 )
 from opensearch_single_kernel.core.relations import (
+    JwtData,
     PeerCluster,
     PeerClusterData,
     PeerClusterOrchestratorData,
@@ -52,6 +56,7 @@ from opensearch_single_kernel.lib.charms.data_platform_libs.v0.data_interfaces i
     DataPeerData,
     DataPeerUnitData,
     OpenSearchProvidesData,
+    SecretGroup,
 )
 from opensearch_single_kernel.utils.helpers import (
     format_unit_name,
@@ -69,7 +74,10 @@ class OpenSearchServer(RelationState):
     """State/Relation data collection for an opensearch unit"""
 
     def __init__(
-        self, relation: Relation | None, data_interface: DataPeerUnitData, component: Unit
+        self,
+        relation: Relation | None,
+        data_interface: DataPeerUnitData,
+        component: Unit,
     ):
         super().__init__(relation, data_interface, component)
         self.unit = component
@@ -232,6 +240,49 @@ class OpenSearchServer(RelationState):
             return
         self.put_object("plugin_config_info", value)
 
+    @property
+    def jwt_auth_configuration(self) -> JWTAuthConfiguration | None:
+        """Return JWT auth configuration if any."""
+        if not (config := self.get_object("jwt-auth-configuration")):
+            return None
+        return JWTAuthConfiguration.from_dict(config)
+
+    @jwt_auth_configuration.setter
+    def jwt_auth_configuration(self, value: JWTAuthConfiguration | None) -> None:
+        """Set or remove JWT auth configuration."""
+        if value is None:
+            self.update({"jwt-auth-configuration": ""})
+        else:
+            self.put_object("jwt-auth-configuration", value.to_dict())
+
+    @property
+    def oauth_openid_connect_url(self) -> str | None:
+        """Return OAuth openid_connect_url if configured."""
+        return self.relation_data.get("oauth_openid_connect_url")
+
+    @oauth_openid_connect_url.setter
+    def oauth_openid_connect_url(self, value: str | None) -> None:
+        """Set or remove OAuth openid_connect_url."""
+        self.update({"oauth_openid_connect_url": value or ""})
+
+    @property
+    def oauth_departing(self) -> bool:
+        """Return whether oauth relation broken event should be skipped.
+
+        When current leader is unit oauth relation isn't breaking
+        even if unit receives oauth relation broken event.
+        """
+        return self.relation_data.get("oauth_departing", "") == "True"
+
+    @oauth_departing.setter
+    def oauth_departing(self, value: bool):
+        """Set whether oauth relation broken event should be skipped.
+
+        When current leader is unit oauth relation isn't breaking
+        even if unit receives oauth relation broken event.
+        """
+        self.update({"oauth_departing": str(value)})
+
 
 class OpenSearchApplication(RelationState):
     """An OpenSearch Application is a charm application with a given role.
@@ -241,7 +292,10 @@ class OpenSearchApplication(RelationState):
     """
 
     def __init__(
-        self, relation: Relation | None, data_interface: DataPeerData, component: Application
+        self,
+        relation: Relation | None,
+        data_interface: DataPeerData,
+        component: Application,
     ):
         super().__init__(relation, data_interface, component)
         self.app = component
@@ -366,7 +420,9 @@ class OpenSearchApplication(RelationState):
     def is_data_role_in_cluster_fleet_apps(self) -> bool:
         """Look for data-role through all the roles of all the nodes in all applications"""
         data_apps_in_fleet = [app for app in self.apps_in_fleet() if "data" in app.roles]
-        return data_apps_in_fleet and any(app.planned_units > 0 for app in data_apps_in_fleet)
+        return bool(data_apps_in_fleet) and any(
+            app.planned_units > 0 for app in data_apps_in_fleet
+        )
 
     @property
     def client_users_dict(self) -> dict[str, str]:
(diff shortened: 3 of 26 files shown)


opensearch_single_kernel/core/state.py
```diff
@@ -211,6 +213,50 @@ def last_host_ip(self, value: str) -> None:
...
+    def jwt_auth_configuration(self, value: JWTAuthConfiguration | None) -> None:
+        """Set or remove JWT auth configuration."""
+        if value is None:
+            self.update({"jwt-auth-configuration": ""})
+        else:
+            self.put_object("jwt-auth-configuration", value.to_dict())
+
+    @property
+    def oauth_openid_connect_url(self) -> str | None:
+        """Return OAuth configuration if any."""
```
akram09: to be clear: Return OAuth openid_connect_url

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+
+class JwtConfigurationRequires(RequirerData):
+    """Data Interface to JWT relation on requirer side."""
+
+    def __init__(self, model, relation_name: str):
+        super().__init__(
+            model,
+            relation_name,
+            additional_secret_fields=["signing-key"],
+        )
```
akram09: Let's have this in relations.py

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+        self.framework.observe(
+            self.charm.on[JWT_CONFIG_RELATION].relation_broken,
+            self._on_jwt_relation_broken,
+        )
+
+    @property
+    def jwt_relation(self) -> ops.Relation | None:
+        """Return the jwt relation if present."""
+        return self.jwt_requires.relations[0] if len(self.jwt_requires.relations) else None
```
akram09: This should be in the state

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+        ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            if self.charm.unit.is_leader():
+                self.charm.status.set(CharmStatuses.JWT_RELATION_INVALID, app=True)
+            return
+
+        if not self.jwt_relation:
+            logger.error(f"Cannot access relation data for {JWT_CONFIG_RELATION}")
+            return
+
+        relation_data = self.jwt_requires.fetch_relation_data([self.jwt_relation.id])
```
akram09: I believe this is our chance to move the jwt data to state. Lets have an object JwtState  where we expose properties and setters for these properties.

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+        jwt_config = relation_data[self.jwt_relation.id]
+
+        self._validate_and_apply_jwt_auth_config(jwt_config)
```
akram09: This can be fetched directly from the state for example

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+        """Handle the removal of the relation."""
+        if (
+            deployment_desc := self.charm.state.application.deployment_desc
+        ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            if self.charm.unit.is_leader():
+                self.charm.status.clear(CharmStatuses.JWT_RELATION_INVALID, app=True)
+            return
+
+        self.charm.state.server.jwt_auth_configuration = None
+        self.charm.config_manager.update_security_config()
```
akram09, line 83: I would call it differently since we have another function with same name lets call this `update_security_yaml_config` and the other on `run_security_admin`

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+        if event.secret.label != self.jwt_requires._generate_secret_label(
+            relation.name,
+            relation.id,
+            "extra",
+        ):
+            logging.debug("Updated secret not relevant")
+            return
+
+        relation_data = self.jwt_requires.fetch_relation_data([self.jwt_relation.id])
+        jwt_config = relation_data[self.jwt_relation.id]
```
akram09: This should move to state

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+        self._validate_and_apply_jwt_auth_config(jwt_config)
+
+    def _validate_and_apply_jwt_auth_config(self, jwt_config: dict[str, str]) -> None:
```
akram09: This function should be decomposed in two functions. I also see that we have duplicated code regarding updating security_config and updating security_index. Let's extract it into a function and call it from validate and from relation broken

opensearch_single_kernel/events/oauth.py
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self, charm: "OpenSearchBaseCharm") -> None:
+        super().__init__(charm, "oauth")
+        self.charm = charm
+
+        # NOTE: Placeholder config options, not really needed by Opensearch
+        client_config = ClientConfig(
+            audience=["opensearch"],
+            redirect_uri="http://opensearch.local",
+            scope="openid email profile",
+            grant_types=["client_credentials"],
```
akram09: Lets have these as constants

opensearch_single_kernel/events/oauth.py
```diff
@@ -0,0 +1,160 @@
...
+            deployment_desc := self.charm.state.application.deployment_desc
+        ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            # in large deployments, OAuth config must only be handled by the main orchestrator
+            # this is a safeguard to avoid different sources for applying security configuration
+            if self.charm.unit.is_leader():
+                self.charm.status.set(CharmStatuses.OAUTH_RELATION_INVALID, app=True)
+            return
+
+        if not (relation := self.model.get_relation(OAUTH_RELATION)):
+            return
```
akram09: Should be in state

opensearch_single_kernel/events/oauth.py
```diff
@@ -0,0 +1,160 @@
...
+
+        if not self.charm.state.application.is_security_index_initialised:
+            logger.debug("Deferring oauth relation changed event as cluster is not ready yet")
+            event.defer()
+            return
+
+        self.charm.state.server.oauth_openid_connect_url = (
+            f"{relation.data[relation.app].get('issuer_url')}/.well-known/openid-configuration"
+        )
+        self.charm.config_manager.update_security_config()
```
akram09, line 111: I like the new reconcile function in config. Makes it very easy and predictable

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,204 @@
...
+            return
+
+        if not self.charm.state.application.is_security_index_initialised:
+            logger.debug("Security index has not been initialized, cannot update security config.")
+            return
+
+        logger.info("Updating security configuration")
+        admin_secrets = self.charm.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
+
+        self.charm.cluster_manager.update_security_config(
```
akram09: We are not handling exception here

opensearch_single_kernel/managers/config.py
```diff
@@ -242,28 +268,208 @@ def _update_static_jvm_options(self) -> None:
...
+    def _security_authc_jwt_config(self) -> dict[str, Any]:
+        """Get set of JWT auth config options for the Opensearch security.
+
+        Intended for authc category in opensearch-security/config.yml config file.
+        """
+        jwt_config = self.state.server.jwt_auth_configuration
+        return {
+            "jwt_auth_domain": {
+                "description": "Authenticate via Json Web Token",
+                "http_enabled": jwt_config is not None,
```
akram09: bool(jwt_config) ?

opensearch_single_kernel/charmcraft.yaml
```diff
@@ -14,3 +14,7 @@ bases:
     run-on:
       - name: "ubuntu"
         channel: "22.04"
+
+charm-libs:
```
akram09, line 18: Why do we need this ?

akram09 (commented):
Thank you Serhii, nice work. A couple of comments from me. Please don't forget to add the oauth integration tests as well.

opensearch_single_kernel/core/relations.py
```diff
@@ -213,6 +214,33 @@ def __bool__(self) -> bool:
         except AttributeError:
             return False
 
+    def get_object(self, key: str) -> dict[str, Any] | None:
+        """Get dict / json object from the relation data store."""
+        data = self.relation_data.get(key)
+        if data is None:
```
Mehdi-Bendriss: nit walrus

opensearch_single_kernel/core/state.py
```diff
@@ -211,6 +213,50 @@ def last_host_ip(self, value: str) -> None:
         """Set the value of last configured IP for the unit. Used for tracking the IP change."""
         self.update({"last_host_ip": value})
 
+    @property
+    def jwt_auth_configuration(self) -> JWTAuthConfiguration | None:
+        """Return JWT auth configuration if any."""
+        config = self.get_object("jwt-auth-configuration")
+        if not config:
```
Mehdi-Bendriss: nit: walrus

opensearch_single_kernel/events/oauth.py
Mehdi-Bendriss: +1

Mehdi-Bendriss (commented):
Thanks Serhii! good work. Not much to add on top of Akram's review.
For confirmation, are all these workflows also manually tested?

opensearch_single_kernel/events/jwt.py
s-mlnk: I renamed `update_security_config` to the `apply_security_config`.

opensearch_single_kernel/events/oauth.py
s-mlnk: It is the only occurrence, should it be moved anyways?

opensearch_single_kernel/events/jwt.py
s-mlnk: We're handling it in both places `_update_security_index` function is called.

opensearch_single_kernel/charmcraft.yaml
s-mlnk: It allows the use of `charmcraft fetch-libs`.

s-mlnk:
~~I will rebase to remove merge conflict warning right after I address all of the PR comments. This way it won't interfere with your reviews.~~
UPD: rebased due to [this request](https://chat.canonical.com/canonical-guest/pl/k7antr61jjnq7csxuqc35woney).

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (dismissed):

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,161 @@
...
+        """Update Opensearch's security index after updating the JWT auth configuration."""
+        if not self.charm.unit.is_leader():
+            return
+
+        if not self.charm.state.application.is_security_index_initialised:
+            logger.debug("Security index has not been initialized, cannot update security config.")
+            return
+
+        logger.info("Updating security configuration")
+        admin_secrets = self.charm.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
```
gatici: We do not handle missing Admin TLS  secrets safely but oauth integration already handles this.We can add  a guard here such as:
```
       if not (
           admin_secrets := self.charm.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
      ):
          event.defer()
          return
```

gatici (commented):
Thanks @s-mlnk. Looks good to me. 
I have one minor comment and I recommend to add unit tests for JWT and Oauth. Besides, there is no integration tests for Oauth.

s-mlnk:
@gatici Thanks for the feedback!
I cannot add OAuth integration tests as it's fully dependent on the client relations that aren't implemented here yet. And also there are no unit tests related to JWT or OAuth in the original repo, so haven't added any too. Moreover, I think there's no much to test anyways, as its only nature is to add a config option to the state and update OpenSearch config.

gatici:
> @gatici Thanks for the feedback! I cannot add OAuth integration tests as it's fully dependent on the client relations that aren't implemented here yet. And also there are no unit tests related to JWT or OAuth in the original repo, so haven't added any too. Moreover, I think there's no much to test anyways, as its only nature is to add a config option to the state and update OpenSearch config.

OK. You are right.

gatici (dismissed):

opensearch_single_kernel/events/jwt.py
s-mlnk: You're right! I've refactored JWT to match OAuth checks.

opensearch_single_kernel/events/jwt.py
```diff
@@ -0,0 +1,154 @@
+# Copyright 2025 Canonical Ltd.
```
akram09: Copyright 2026

opensearch_single_kernel/events/oauth.py
akram09: Yes

akram09 (commented):
Seems good for me. Lets agree on adding the `test_oauth.py` integration test after merging the opensearch-client PR

opensearch_single_kernel/core/state.py
```diff
@@ -232,6 +240,49 @@ def plugin_config_info(self, value: dict[str, PluginConfigInfo]) -> None:
...
+            return None
+        return JWTAuthConfiguration.from_dict(config)
+
+    @jwt_auth_configuration.setter
+    def jwt_auth_configuration(self, value: JWTAuthConfiguration | None) -> None:
+        """Set or remove JWT auth configuration."""
+        if value is None:
+            self.update({"jwt-auth-configuration": ""})
+        else:
+            self.put_object("jwt-auth-configuration", value.to_dict())
```
akram09: This is not good for consistency . Lets have a deleter and use `del` for removing the properties what do you think ?

opensearch_single_kernel/events/jwt.py
akram09: This is not fixed !

opensearch_single_kernel/events/oauth.py
```diff
@@ -0,0 +1,169 @@
+# Copyright 2025 Canonical Ltd.
```
akram09: Copyright as well

akram09 (dismissed):
Seems good for me

opensearch_single_kernel/core/state.py
s-mlnk: Good catch!

gatici (approved):
LGTM

reneradoi (approved):
Did a high-level review, no concerns from my side.

akram09 (approved):
Seems good 👍

PR #30, opened 2026-02-27
patch: Adding tics workflow

akram09:
This PR adds the TICS Github workflow to the OpenSearch Single Kernel Repository. For the reference, a new project has been created named 'opensearch-single-kernel' in the TICS dashboard. This workflow will use the feature/single-kernel for now until the refactoring is done and we merge to 2/edge. The workflow that is being added is the same previously used for [opensearch-operator](https://github.com/canonical/opensearch-operator/) repository.

### DIFF
--- a/.github/workflows/tics_run.yaml
+++ b/.github/workflows/tics_run.yaml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   tics-scan:
-    runs-on: [self-hosted, noble, tiobe]
+    runs-on: [self-hosted, linux, amd64, tiobe, noble]
 
     steps:
       - name: Checkout the project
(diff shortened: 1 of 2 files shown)


.github/workflows/tics_run.yaml
```diff
@@ -0,0 +1,61 @@
+name: TICS run self-hosted test
+
+on:
+    schedule:
+        - cron: '0 2 * * 6' # Every Saturday 2:00 AM UTC
+    workflow_dispatch: # Allows manual triggering
+
+jobs:
+  tics-scan:
+    runs-on: [self-hosted, noble, tiobe]
```
Mehdi-Bendriss:
```suggestion
    runs-on: [self-hosted, linux, amd64, tiobe, noble]
```

Mehdi-Bendriss (approved):

PR #33, opened 2026-03-04
patch: [DPE-9402] COS

s-mlnk:
# Key changes

* Ported COS into the single COS events handler due to use of multiple managers, however no events handling actually occurs there, only configuration.
* Added COS Agent charm library.
* Copied Prometheus alerts config file.
* Copied Grafana dashboard config file.

These changes verified with both `grafana-agent` `1/stable` and `2/edge`.

### DIFF
--- a/opensearch_single_kernel/events/cos.py
+++ b/opensearch_single_kernel/events/cos.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Handler for COS configuration & events."""
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+from ops import Object
+
+from opensearch_single_kernel.common.constants import (
+    COS_PORT,
+    COS_RELATION,
+    COS_USER,
+    PEER_CLUSTER_RELATION,
+    PEER_RELATION,
+    CertType,
+    Scope,
+)
+from opensearch_single_kernel.lib.charms.grafana_agent.v0.cos_agent import (
+    COSAgentProvider,
+)
+from opensearch_single_kernel.utils.secrets import password_key
+
+if TYPE_CHECKING:
+    from opensearch_single_kernel.charms.base import OpenSearchBaseCharm
+
+logger = logging.getLogger(__name__)
+
+
+class CosEventsHandler(Object):
+    """Class implementing OpenSearch COS configuration & event handling."""
+
+    def __init__(self, charm: "OpenSearchBaseCharm"):
+        super().__init__(charm, key="cos_events")
+        self.charm = charm
+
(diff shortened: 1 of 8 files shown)


opensearch_single_kernel/events/cos.py
```diff
@@ -0,0 +1,97 @@
...
+        self.cos_integration = COSAgentProvider(
+            self.charm,
+            relation_name=COS_RELATION,
+            scrape_configs=self.scrape_config,
+            refresh_events=[
+                self.charm.on.config_changed,
+                self.charm.on.set_password_action,
+                self.charm.on.secret_changed,
+                self.charm.on[PEER_RELATION].relation_changed,
+                self.charm.on[PEER_RELATION].relation_changed,
```
Mehdi-Bendriss: This should be `PERR_CLUSTER_RELATION`

opensearch_single_kernel/events/cos.py
```diff
@@ -0,0 +1,97 @@
...
+            ],
+            metrics_rules_dir="./src/alert_rules/prometheus",
+            log_slots=["opensearch:logs"],
+        )
+
+    def scrape_config(self) -> list[dict[str, Any]]:
```
Mehdi-Bendriss, line 54: Should scrape_config and _get_prometheus_labels not be in the manager?

Mehdi-Bendriss (commented):
Thanks Serhii! I only have a comment and a question/suggestion

s-mlnk:
~~Will rebase to fix merge conflict after all PR comments are addressed.~~
UPD: rebased due to [this request](https://chat.canonical.com/canonical-guest/pl/k7antr61jjnq7csxuqc35woney).

opensearch_single_kernel/events/cos.py
s-mlnk: While I moved `get_prometheus_labels` into the cluster manager, `scrape_config` cannot be moved into a single manager as it uses both cluster & tls managers and it will cause a [managers overlap issue](https://github.com/canonical/opensearch-dashboards-single-kernel-library/pull/2#discussion_r2864087287).

opensearch_single_kernel/events/cos.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
```
akram09: Copyright 2026

akram09 (dismissed):
Seems good for me, thanks @s-mlnk . Please don't forget to add the unit test `helper_cos.py`. And I was wondering why we don't have an integration test for this @Mehdi-Bendriss do you know ?

Mehdi-Bendriss (approved):

gatici (approved):
Looks good. It would be better if we also add some tests.

akram09 (approved):

PR #34, opened 2026-03-05
patch: Adding SSL Adapter to not pass chain as file path 

akram09:
This Pull Request address the issue of having to pass the chain.pem path to the request each time we are calling the opensearch api. The drawback of that approach is that we had to maintain the chain.pem file whenever we are doing ca_rotation or are just creating the certificates. Another drawback is that in kubernetes the filesystem of the workload and the charm is not the same so we would need to cache the chain.pem in the charm container and make sure it is up to date. 

To address this issue the solution is to use an adapter and pass the certificate chain directly from the secrets instead of creating a file and passing the file path.

### DIFF
--- a/opensearch_single_kernel/common/client.py
+++ b/opensearch_single_kernel/common/client.py
@@ -24,7 +24,7 @@ from tenacity.wait import WaitBaseT
 
 from opensearch_single_kernel.common.exceptions import OpenSearchHttpError
 from opensearch_single_kernel.core.models import App, Node
-from opensearch_single_kernel.workload.base import BaseWorkload
+from opensearch_single_kernel.utils.requests import SSLStringAdapter
 
 logger = logging.getLogger(__name__)
 
@@ -33,7 +33,7 @@ class OpenSearchClient:
     """Handle OpenSearch Interaction with Server."""
 
     def __init__(
-        self, workload: BaseWorkload, host: str, port: int, admin_secret: str | None = None
+        self, host: str, port: int, admin_secret: str | None = None, ca_chain: str | None = None
     ):
         """Initialise the client.
 
@@ -41,8 +41,8 @@ class OpenSearchClient:
         """
         self.host = host
         self.port = port
-        self.workload = workload
         self.admin_secret = admin_secret
+        self.ca_chain = ca_chain
 
     def flush_translog(self, alt_hosts: list[str] | None = None) -> None:
         """Flush the OpenSearch translog to ensure all operations are committed to disk."""
@@ -403,16 +403,11 @@ class OpenSearchClient:
         """Call the /_nodes API endpoint of opensearch"""
         return self.request("GET", "/_nodes", host=host, alt_hosts=alt_hosts, retries=3)
 
-    def is_node_up(self, host: str | None = None) -> bool:
-        """Get status of node.
+    def is_opensearch_api_up(self, host: str) -> bool:
+        """Get status of OpenSearch API
 
-        This assumes OpenSearch is Running. Defaults to this unit
+        This assumes OpenSearch is Reachable. Defaults to this unit
         """
-        # This function needs to give us a quick response
-        host = host or self.host
-        if not self.workload.is_reachable(host, self.port):
-            return False
-
         try:
             resp_code = self.request(
                 "GET",
@@ -480,11 +475,13 @@ class OpenSearchClient:
                         s.cert = cert_files
                     else:
                         s.auth = ("admin", self.admin_secret)
-                    # TODO: Handle this when implementing the k8s version of start workflow.
+
+                    if self.ca_chain:
+                        s.mount(url, SSLStringAdapter(self.ca_chain))
+
                     request_kwargs = {
                         "method": method.upper(),
                         "url": url,
-                        "verify": f"{self.workload.paths.certs}/chain.pem",
                         "headers": {
                             "Accept": "application/json",
                             "Content-Type": "application/json",
@@ -519,7 +516,7 @@ class OpenSearchClient:
 
         urls = []
         for host_candidate in (host or self.host, *(alt_hosts or [])):
-            if check_hosts_reach and not self.is_node_up(host_candidate):
+            if check_hosts_reach and not self.is_opensearch_api_up(host_candidate):
                 continue
             urls.append(f"https://{host_candidate}:{self.port}/{endpoint}")
         if not urls:
--- a/opensearch_single_kernel/events/opensearch.py
+++ b/opensearch_single_kernel/events/opensearch.py
@@ -153,7 +153,7 @@ class OpenSearchEventsHandler(Object):
             event.defer()
             return
 
-        if is_node_up := self.charm.cluster_manager.opensearch_client.is_node_up():
+        if is_node_up := self.charm.cluster_manager.is_node_up():
             health = self.charm.status.apply_health(app=self.charm.unit.is_leader())
 
             if health in [HealthColors.UNKNOWN, HealthColors.YELLOW_TEMP]:
@@ -219,7 +219,7 @@ class OpenSearchEventsHandler(Object):
         if not (self.charm.unit.is_leader() and len(event.relation.units) > 0):
             return
 
-        if not self.charm.cluster_manager.opensearch_client.is_node_up():
+        if not self.charm.cluster_manager.is_node_up():
             logger.debug("Node is not up. Deferring event.")
             event.defer()
             return
@@ -294,8 +294,7 @@ class OpenSearchEventsHandler(Object):
                     raise OpenSearchHAError(CharmStatuses.CLUSTER_HEALTH_RED.value.message)
         finally:
             if planned_units > 1 and (
-                self.charm.cluster_manager.opensearch_client.is_node_up()
-                or self.charm.cluster_manager.alt_hosts
+                self.charm.cluster_manager.is_node_up() or self.charm.cluster_manager.alt_hosts
             ):
                 # release lock
                 self.charm.lock_manager.release()
@@ -319,7 +318,7 @@ class OpenSearchEventsHandler(Object):
             return
 
         # if node already shutdown - leave
-        if not self.charm.cluster_manager.opensearch_client.is_node_up():
+        if not self.charm.cluster_manager.is_node_up():
             return
 
         # review available CMs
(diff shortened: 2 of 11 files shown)


opensearch_single_kernel/common/client.py
```diff
@@ -403,16 +403,11 @@ def get_nodes(self, host: str | None = None, alt_hosts: list[str] | None = None)
...
+    def is_opensearch_api_up(self, host: str) -> bool:
+        """Get status of OpenSearch API
 
-        This assumes OpenSearch is Running. Defaults to this unit
+        This assumes OpenSearch is Reachable. Defaults to this unit
         """
-        # This function needs to give us a quick response
-        host = host or self.host
-        if not self.workload.is_reachable(host, self.port):
-            return False
```
Mehdi-Bendriss, line 418: no longer needed?

opensearch_single_kernel/common/client.py
```diff
@@ -403,16 +403,11 @@ def get_nodes(self, host: str | None = None, alt_hosts: list[str] | None = None)
         """Call the /_nodes API endpoint of opensearch"""
         return self.request("GET", "/_nodes", host=host, alt_hosts=alt_hosts, retries=3)
 
-    def is_node_up(self, host: str | None = None) -> bool:
-        """Get status of node.
+    def is_opensearch_api_up(self, host: str) -> bool:
```
Mehdi-Bendriss, line 406: I'd say having `api_up` in the name is a bit confusing, implies that there are multiple definitions to `up`.
`is_node_up` is more clear here.

opensearch_single_kernel/events/opensearch.py
```diff
@@ -219,7 +219,7 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
         if not (self.charm.unit.is_leader() and len(event.relation.units) > 0):
             return
 
-        if not self.charm.cluster_manager.opensearch_client.is_node_up():
```
Mehdi-Bendriss, line 222: is it not better to keep `is_node_up` in the client instead of the cluster_manager, since it can potentially be used by every manager/event handler out there

opensearch_single_kernel/managers/tls.py
```diff
@@ -469,12 +450,6 @@ def reload_tls_certificates(self) -> None:
                 cert_files=(str(tmp_cert), str(tmp_key))
             )
 
-    def finalize_ca_certs_rotation(self) -> None:
```
Mehdi-Bendriss, line 472: nice

opensearch_single_kernel/utils/requests.py
```diff
@@ -0,0 +1,31 @@
...
+
+class SSLStringAdapter(HTTPAdapter):
+    """Custom Transport Adapter that loads SSL certificates directly from a string."""
+
+    def __init__(self, cert_string, **kwargs):
+        """Initialize the adapter with the certificate string."""
+        self.cert_string = cert_string
+        super().__init__(**kwargs)
+
+    def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs):
```
Mehdi-Bendriss:
```suggestion
    def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs) -> None:
```

Mehdi-Bendriss (commented):
Thanks Akram! good work. Though many TLS tests are failing, we should investigate, as a lot of this PR has been touching on the TLS flows

opensearch_single_kernel/common/client.py
akram09: What I have done here is that I separated the is_node_up into two function one function in the client that checks if the API is responding to requests. and the main function `is_node_up` in the BaseManager which checks first if it is reachable and then check if the api is up. with this I have removed a dependency from the client to the workload.

opensearch_single_kernel/events/opensearch.py
akram09: It is in the BaseManager now and it will be available to all the managers.  For event handler they can access it through the cluster manager. I have done this free the opensearch-client from dependency on the workload.

opensearch_single_kernel/utils/requests.py
```diff
@@ -0,0 +1,31 @@
...
+        """Initialize the pool manager with a custom SSL context."""
+        context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        context.verify_mode = ssl.CERT_REQUIRED
+        context.check_hostname = True
+        try:
+            context.load_verify_locations(cadata=self.cert_string)
+        except Exception as e:
```
reneradoi, line 26: This could be reduced to `ssl.SSLError` I believe.

opensearch_single_kernel/utils/requests.py
```diff
@@ -0,0 +1,31 @@
...
+    def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs) -> None:
+        """Initialize the pool manager with a custom SSL context."""
+        context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        context.verify_mode = ssl.CERT_REQUIRED
+        context.check_hostname = True
+        try:
+            context.load_verify_locations(cadata=self.cert_string)
+        except Exception as e:
+            print(f"Error loading certificate string: {e}")
+            raise
```
reneradoi, line 28: If we let the error bubble up, this could crash in case of malformatted cert strings. Currently neither `Exception` nor `SSLError` (as I suggested) is being caught.

opensearch_single_kernel/utils/requests.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs) -> None:
+        """Initialize the pool manager with a custom SSL context."""
+        context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        context.verify_mode = ssl.CERT_REQUIRED
+        context.check_hostname = True
+        try:
+            context.load_verify_locations(cadata=self.cert_string)
+        except Exception as e:
+            print(f"Error loading certificate string: {e}")
```
reneradoi, line 27: Why print?

opensearch_single_kernel/common/client.py
```diff
@@ -484,11 +475,13 @@ def call(urls: list[str]) -> requests.Response:
                         s.cert = cert_files
                     else:
                         s.auth = ("admin", self.admin_secret)
-                    # TODO: Handle this when implementing the k8s version of start workflow.
+
+                    if self.ca_chain:
```
reneradoi, line 479: This seems to be a behavioral change: While we always use the `chain.pem` currently and fail if that is not there, we now allow for non-verified connections in cases where the admin secret is not populated or the `ca_chain` is `None` (for whatever reason). Is this intended? Wouldn't it be better to raise in this case?

opensearch_single_kernel/managers/base.py
```diff
@@ -28,9 +32,15 @@ def __init__(self, state: ClusterState, workload: BaseWorkload):
...
+        # Get chain.pem from secrets and pass it to client
+        admin_secret = self.state.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val, peek=True)
+        ca_chain = admin_secret.get("chain")
         return OpenSearchClient(
-            self.workload, self.state.host_ip, OPENSEARCH_HTTP_PORT, admin_secret
+            self.state.host_ip,
+            OPENSEARCH_HTTP_PORT,
+            admin_password,
+            ca_chain=ca_chain,
         )
```
reneradoi, line 45: From a conceptual point of view, I think this might cause issues on the CA rotation workflow. While we previously updated the `chain.pem` on disk after certain checkpoints in the workflow, we now use the new CA cert immediately after updating the admin secret in the `certificiate_availabable` event (`self.charm.tls_manager.update_certificate_secret_if_needed` in line 194). That means, if the CA was updated, the operator will not be able to verify the CA that Opensearch presents anymore.

reneradoi (changes requested):
Hey Akram! While I really like the idea behind this change, I think it does have a conceptual flaw currently, which is the case where the CA in the secret is not the same as Opensearch currently uses. This needs to be addressed, because it breaks the CA rotation workflow.

While the previous solution relied on a file stored on disk, the architectural idea behind this was to only update the file once Opensearch was updated to use the new CA (intended delay from the certificate available with the new CA in the cert). This is now not possible anymore, or the secret would have to be duplicated somehow for the time until the new CA was loaded into Opensearch.

I don't have a clear idea yet for how to solve this though with this new approach.

opensearch_single_kernel/utils/requests.py
reneradoi: Sorry, I have to correct myself here, as I just saw that `Exception` is actually caught in `OpensearchClient.request()`.

opensearch_single_kernel/managers/tls.py
```diff
@@ -306,25 +306,6 @@ def is_secret_found(secrets: dict[str, str] | None) -> bool:
...
-        # we store the pem format to make it easier for the python requests lib
-        chain_path = self.workload.paths.certs / "chain.pem"
-        if parent_dir_path := chain_path.parent:
-            self.workload.mkdir(parent_dir_path, parents=True, exist_ok=True)
-
-        # if the chain.pem already contains the current CA chain, we can skip rewriting it
-        bundle_content = self.workload.read_text(chain_path) if chain_path.exists() else ""
-        if ca_chain not in bundle_content:
-            self.workload.write_text(f"{bundle_content}\n{ca_chain}", chain_path)
-
```
gatici, line 327: The new approach removes the mixed old+new CA bundle that was previously used during CA rotation. With this change, the client now trusts only the current chain from secrets, while the old transition bundle logic is gone. During rotation, some nodes may still present certificates signed by the old CA, so charm/workload HTTPS calls to those nodes can fail.

This transition state is also no longer durable. The old CA is not available through secrets anymore, and keeping it only in memory would not survive a VM or pod restart. That is especially problematic on k8s, where the filesystem is ephemeral and secrets are effectively the source of truth: if a pod restarts mid-rotation, there is no reliable way to reconstruct the old+new CA overlap, so TLS connectivity to nodes still using the old CA can break.
I do not see a reliable alternative here other than persisting the old CA for the duration of the rotation, which is effectively what the previous implementation was.

PR #46, opened 2026-03-24
patch: [DPE-9485] Lock manager refactor

s-mlnk:
### Key changes
* All Lock manager requests moved to OpenSearch Client.
* Whole Lock manager state manipulation (of lock relation) is moved to state's `LockAppState` & `LockServerState` along with some of related logic.
* `LockManager.acquired` property was renamed to `LockManager.acquire()` function to avoid confusion of intent.


opensearch_single_kernel/managers/lock.py
```diff
@@ -293,14 +196,14 @@ def acquired(self) -> bool:  # noqa: C901
             logger.debug("[Node lock] Opensearch %s", online_nodes)
             assert online_nodes > 0
             try:
-                unit = self.unit_with_lock(host)
+                unit = self.opensearch_client.get_unit_with_lock(host, self.alt_hosts)
```
reneradoi: nit: `unit` could be renamed to `unit_with_lock` for better readability.

reneradoi (dismissed):
No concerns from my side, just a minor nit.

opensearch_single_kernel/common/client.py
```diff
@@ -895,3 +896,153 @@ def log_error(retry_state: RetryCallState):
...
+                logger.error("Failed to write OpenSearch lock document to all nodes.")
+                logger.debug(
+                    "[Node lock] Deleting OpenSearch lock after failing to write to all nodes"
+                )
+                # Delete document id 0
+                self.delete_lock_document(host, alt_hosts)
+                logger.debug(
+                    "[Node lock] Deleted OpenSearch lock after failing to write to all nodes"
+                )
+                return False
```
akram09, line 1491: I think we had a discussion with Mehdi, on doing some research to improve the efficiency of this request. @Mehdi-Bendriss do we want to do that in this PR ?

opensearch_single_kernel/core/state.py
```diff
@@ -582,6 +584,99 @@ def auth_configuration(self) -> JWTAuthConfiguration:
...
+        """Get format name of the unit that holds the lock."""
+        return self.relation_data.get("unit-with-lock")
+
+    @unit_with_lock.setter
+    def unit_with_lock(self, value: str) -> None:
+        """Set format name of the unit that holds the lock.
+
+        Update leader_acquired_after_juju_event_id if lock acquired by the current (leader) unit.
+        """
+        assert self.unit_with_lock != value
```
akram09, line 631: What would happen when this fails, I don't see any handling of assertError. We might need to raise a custom exception here

opensearch_single_kernel/core/state.py
```diff
@@ -582,6 +584,99 @@ def auth_configuration(self) -> JWTAuthConfiguration:
...
+        data_interface: Data,
+        component: Unit,
+    ):
+        super().__init__(relation, data_interface, component)
+        self.unit = component
+
+    @property
+    def lock_requested(self) -> bool:
+        """Get whether the lock is requested by unit."""
+        return self.relation_data.get("lock_requested", "") == "True"
```
akram09: We will use lower case for this

opensearch_single_kernel/core/state.py
```diff
@@ -1032,3 +1127,55 @@ def jwt(self) -> JwtState:
     def oauth_relation(self) -> Relation | None:
         """Get OAuth relation."""
         return self.model.get_relation(OAUTH_RELATION)
+
+    @property
+    def lock_server(self) -> LockServerState:
```
akram09: nit: server_lock

opensearch_single_kernel/core/state.py
```diff
@@ -1032,3 +1127,55 @@ def jwt(self) -> JwtState:
...
+                    component=unit,
+                )
+                for unit in (self.server.unit, *self.lock_relation.units)
+            ]
+            if self.lock_relation
+            else []
+        )
+
+    @property
+    def lock_application(self) -> LockAppState:
```
akram09: nit: application_lock

opensearch_single_kernel/managers/lock.py
```diff
@@ -102,98 +98,31 @@ def acquired(self) -> bool:
 
     def release(self) -> None:
         """Release lock for this unit."""
-        if not self._relation:
+        if not self.state.lock_relation:
             return
 
-        self._relation.data[self.state.server.unit].pop("lock-requested", None)
+        self.state.lock_server.lock_requested = False
```
akram09: This changes the UX a bit since it will not delete the value from the databag

opensearch_single_kernel/managers/lock.py
```diff
@@ -243,7 +165,7 @@ def should_ignore_lock(self, deployment_desc: DeploymentDescription) -> bool:
                 or (
                     # in case all data-nodes are powered down after being previously started
                     # ignore the lock to get a data-node started, as it holds security index
-                    self.state.server.started
+                    bool(self.state.server.started)
```
akram09, line 172: Nice catch

akram09 (dismissed):
Great work Serhii !

opensearch_single_kernel/common/client.py
s-mlnk: I just checked if those requests are sufficient to keep lock manager working, but haven't thought of any better way.

opensearch_single_kernel/managers/lock.py
s-mlnk: It's only a binary flag. Do we need 2 different methods of doing the same thing? I mean we can reset this flag either with using deleter or setter (like here).

opensearch_single_kernel/core/state.py
s-mlnk: If you're talking about `"True"` check, that is how it's done all across the state.

opensearch_single_kernel/managers/lock.py
s-mlnk: UPD: in call we agreed to use setter but for False values we should remove field from the relation bag instead of writing negative value.

opensearch_single_kernel/core/state.py
akram09: We can update them to be all the same.  `.lower() == "true"`

opensearch_single_kernel/common/client.py
```diff
@@ -895,3 +896,153 @@ def log_error(retry_state: RetryCallState):
...
+        except OpenSearchHttpError as e:
+            if (
+                e.response_code == 400
+                and e.response_body.get("error", {}).get("type")
+                == "resource_already_exists_exception"
+            ):
+                # Index already created
+                return True
+            else:
+                logger.exception("Error creating OpenSearch lock index")
```
Mehdi-Bendriss: I think this should be `logger.error`

opensearch_single_kernel/common/client.py
```diff
@@ -895,3 +896,153 @@ def log_error(retry_state: RetryCallState):
...
+                    )
+                return True
+        except OpenSearchHttpError:
+            pass
+
+        # Create index if it doesn't exist
+        try:
+            self.request(
+                "PUT",
+                endpoint=f"/{OPENSEARCH_LOCK_INDEX}?wait_for_active_shards=all",
```
Mehdi-Bendriss: Could we also add `&refresh=true` to ensure the doc is searchable immediately?

opensearch_single_kernel/common/client.py
Mehdi-Bendriss: I think we are good for the time being, we will reiterate on it using a `_seq_no` and `_primary_term`

Mehdi-Bendriss (commented):
Thanks Serhii, great work! I only have a couple of small comments

akram09 (commented):
Seems good for me, please just add the lock manager unit tests

Mehdi-Bendriss (dismissed):

opensearch_single_kernel/common/client.py
s-mlnk: It leads to the next error:

```
unit-opensearch-2: 18:10:11 ERROR unit.opensearch/2.juju-log Could not create OpenSearch lock index: HTTP error self.response_code=400
self.response_body={'error': {'root_cause': [{'type': 'illegal_argument_exception', 'reason': 'request [/.charm_node_lock] contains unrecognized parameter: [refresh]'}], 'type': 'illegal_argument_exception', 'reason': 'request [/.charm_node_lock] contains unrecognized parameter: [refresh]'}, 'status': 400}
```

So I reverted it.

s-mlnk: I've added a separate refresh operation just after index creation.

Mehdi-Bendriss (approved):
Thanks Serhii!

akram09 (approved):

PR #55, opened 2026-04-06
patch: [DPE-9606] Advanced statuses

s-mlnk:
Key points:
* All of the charm statuses migrated to advanced status lib.
* General events are tied to the cluster manager component.
* Proxy methods `add_status_if_not_present` & `remove_status_if_present` are implemented in state to support interpolated statuses.
* Unit tests are reworked to check on `add_status_if_not_present` usage and not on the computed unit status.


opensearch_single_kernel/common/statuses.py
```diff
@@ -8,147 +8,229 @@
...
+    INSTALL_IN_PROGRESS = StatusObject(
+        status="maintenance", message="Installing OpenSearch...", running="blocking"
+    )
+    SECURITY_INDEX_INIT_IN_PROGRESS = StatusObject(
+        status="maintenance",
+        message="Initializing the security index...",
+        running="blocking",
+    )
+    WAITING_TO_START = StatusObject(status="waiting", message="Waiting for OpenSearch to start...")
+    SERVICE_START_ERROR = StatusObject(
```
akram09, line 29: I am not sure this is should be a running status

opensearch_single_kernel/charms/base.py
```diff
@@ -102,6 +99,19 @@ def __init__(self, *args):
         self.jwt_events = JWTEventsHandler(self)
         self.oauth_events = OAuthEventsHandler(self)
 
+        self.status_handler = StatusHandler(
+            self,
+            self.profiles_manager,
+            self.cluster_manager,
+            self.internal_users_manager,
```
akram09: I think in terms of order. I would advocate for 
- profiles
- health
- cluster 
- lock
- snapshots
- internal_users
- external_clients
- notifications

would want to hear from @Mehdi-Bendriss as well

opensearch_single_kernel/charms/base.py
```diff
@@ -146,7 +160,25 @@ def stop_opensearch(self, *, restart: bool = False) -> None:
...
+    def apply_health(
+        self,
+        wait_for_green_first: bool = False,
+        use_localhost: bool = True,
+        app: bool = True,
+        unit: bool = True,
+    ):
+        """Fetch cluster health and set it on the app status."""
+        if app and not self.unit.is_leader():
+            self.trigger_peer_rel_changed(on_other_units=True)
```
akram09, line 192: Was there a comment to explain this ?

opensearch_single_kernel/common/client.py
```diff
@@ -1006,7 +1006,7 @@ def get_allocation_explain(
         )
 
     def get_health(
-        self, host: str, wait_for_green: bool, alt_hosts: list[str] | None = None
+        self, host: str | None, wait_for_green: bool, alt_hosts: list[str] | None = None
```
akram09, line 1031: Was there a case where we pass this as None ?

opensearch_single_kernel/common/statuses.py
```diff
@@ -8,147 +8,229 @@
...
+    ACTIVE_IDLE = StatusObject(status="active", message="")
+    INSTALL_IN_PROGRESS = StatusObject(
+        status="maintenance", message="Installing OpenSearch...", running="blocking"
+    )
+    SECURITY_INDEX_INIT_IN_PROGRESS = StatusObject(
+        status="maintenance",
+        message="Initializing the security index...",
+        running="blocking",
+    )
+    WAITING_TO_START = StatusObject(status="waiting", message="Waiting for OpenSearch to start...")
```
akram09: Shouldn't this be a running status ?

opensearch_single_kernel/common/statuses.py
```diff
@@ -8,147 +8,229 @@
...
+        status="blocked",
+        message="An error occurred during the start of the OpenSearch service.",
+        running="async",
+    )
+    SERVICE_IS_STOPPING = StatusObject(
+        status="waiting",
+        message="The OpenSearch service is stopping.",
+        running="blocking",
+    )
+    OAUTH_RELATION_INVALID = StatusObject(
```
akram09: OAUTH and JWT should be in a separate Enum

opensearch_single_kernel/common/statuses.py
```diff
@@ -8,147 +8,229 @@
...
-    SECURITY_INDEX_UPDATE_ERROR = BlockedStatus(
-        "Failed to update security configuration, check logs for details."
-    )
+class LockStatuses(Enum):
+    """Collection of charm statuses related to lock manager."""
 
-    # Start
-    SERVICE_START_ERROR = BlockedStatus(
-        "An error occurred during the start of the OpenSearch service."
+    REQUEST_LOCK_ON_START = StatusObject(
```
akram09, line 99: I think lock requesting should be a running status

opensearch_single_kernel/core/state.py
```diff
@@ -144,6 +160,15 @@ def started(self) -> str:
...
         return self.relation.data[self.unit].get("started", "")
 
+    @started.setter
+    def started(self, value: str) -> None:
+        """Get the value of 'started' key from unit data bag"""
+        self.relation_data.update({"started": value})
+
+    @started.deleter
+    def started(self) -> None:
+        self.relation_data.update({"started": ""})
```
akram09: Just so that we don't have annoying warnings. Add an if check to the deleters where we check if the property exists first before deleting.

opensearch_single_kernel/core/state.py
```diff
@@ -166,7 +191,11 @@ def tls_ca_renewed(self, value: bool):
 
     @property
     def tls_configured(self) -> bool:
-        """Get the value of 'tls_configured' from unit data bag."""
+        """Get the value of 'tls_configured' from unit data bag.
+
+        This property signalizes the completion of TlsManager.is_fully_configured()
```
akram09: "This property signal..."

opensearch_single_kernel/core/state.py
```diff
@@ -798,13 +827,19 @@ def __init__(self, charm: "OpenSearchBaseCharm", substrate: Substrates):
...
 
+        self.statuses = StatusesState(self, STATUS_PEERS_RELATION)
         # TODO: Add secrets
         self.peer_app_interface = DataPeerData(model=charm.model, relation_name=PEER_RELATION)
         self.peer_unit_interface = DataPeerUnitData(model=charm.model, relation_name=PEER_RELATION)
         self.client_data_interface = OpenSearchProvidesData(
             model=charm.model, relation_name=CLIENT_RELATION
         )
 
+        self.smtp_requires = SmtpRequires(charm, SMTP_RELATION)
```
akram09: I think we need to add the missing ones like jwt and oauth no ?

 Also I am not sure it is the right way to add the requirers here in the state I would advocate for adding them to the charm and injecting what we need since we don't want to have event handling in state.

opensearch_single_kernel/core/state.py
```diff
@@ -1304,3 +1338,140 @@ def application_lock(self) -> LockAppState:
             component=self.model.app,
             unit_name=self.unit_name,
         )
+
+    @property
+    def smtp_relations(self) -> list[Relation]:
+        """Get SMTP relations."""
```
akram09: Please add it on top where we have the other relations just so we follow the order

opensearch_single_kernel/core/state.py
```diff
@@ -1304,3 +1338,140 @@ def application_lock(self) -> LockAppState:
...
+        search_parameters: dict[str, Any] | None = None,
+    ) -> None:
+        """Add charm status if not present already.
+
+        Args:
+            status: charm status to be added.
+            scope: scope of the added charm status.
+            component: name of the responsible component of the added status.
+            dynamic_params: params to format added status message with.
+            search_parameters: params to format searched status message with prior to interpolated
```
akram09, line 1375: I see what you want to do here. two points
- change parameter name to `override_parameters`
- Why not having None instead of "{}" ?

opensearch_single_kernel/core/state.py
```diff
@@ -1304,3 +1338,140 @@ def application_lock(self) -> LockAppState:
...
+            # Updates dynamic params if status already present.
+            self.remove_status_if_present(status, scope, component, interpolated=True)
+            self.statuses.add(format_status(status, dynamic_params), scope, component)
+
+    def remove_status_if_present(
+        self,
+        status: StatusObject,
+        scope: AdvancedStatusesScope,
+        component: str,
+        interpolated: bool = False,
```
akram09, line 1410: I don't see why we need an "interpolated" argument we could basically just check if the "search_parameters" is not None .

opensearch_single_kernel/core/state.py
```diff
@@ -1304,3 +1338,140 @@ def application_lock(self) -> LockAppState:
...
+                if interpolated_parameters
+                else status
+            ).message,
+        )
+        for present_status in self.statuses.get(scope, component):
+            if re.fullmatch(regex_pattern, present_status.message) is not None:
+                return present_status
+        return None
+
+    def get_storage_connection_info_from_relation(
```
akram09, line 1470: This is very helpful. Thanks for adding it I will benefit a lot from it in my LD PR

opensearch_single_kernel/events/external_clients.py
```diff
@@ -78,40 +78,78 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
             return
 
         if not validate_index_name(event.index):
-            self.charm.status.set(
-                CharmStatuses.INVALID_INDEX_NAME, dynamic_params={"index": event.index}
+            self.charm.status_handler.set_running_status(
```
akram09: why not use add_status function from state ?

opensearch_single_kernel/events/jwt.py
```diff
@@ -58,7 +58,16 @@ def _on_jwt_relation_created(self, _: RelationCreatedEvent) -> None:
             # in large deployments, JWT configuration must only be handled by the main orchestrator
             # this is a safeguard to avoid different sources for applying security configuration
             if self.charm.unit.is_leader():
-                self.charm.status.set(CharmStatuses.JWT_RELATION_INVALID, app=True)
+                self.charm.state.add_status_if_not_present(
+                    GeneralStatuses.JWT_RELATION_INVALID.value,
+                    "app",
+                    self.charm.cluster_manager.name,
+                )
+                self.charm.state.remove_status_if_present(
```
akram09: is this needed ?

opensearch_single_kernel/events/jwt.py
```diff
@@ -140,11 +175,7 @@ def apply_security_config_if_needed(self, event: EventBase) -> None:
                 admin_secrets, self.charm.config_manager.SECURITY_CONFIG_YML
             )
             logger.info("Updated Opensearch security index")
-        except OpenSearchCmdError as e:
-            logger.debug(f"Error when updating the security index: {e.out}")
-            self.charm.status.set(CharmStatuses.SECURITY_INDEX_UPDATE_ERROR, app=True)
+        except OpenSearchCmdError:
```
akram09: Why removed status setting here ?

opensearch_single_kernel/events/notifications.py
```diff
@@ -219,19 +281,22 @@ def _on_smtp_credentials_gone(self, event: RelationBrokenEvent) -> None:  # noqa
...
+                    search_parameters={"id": event.relation.id},
+                )
+            if not self.charm.state.smtp_relations:
+                self.charm.state.remove_status_if_present(
+                    NotificationsStatuses.SMTP_RELATION_INVALID.value,
+                    "app",
+                    self.charm.notifications_manager.name,
+                )
 
         label = self.charm.notifications_manager.label(event.relation.id)
```
akram09, line 294: I think the logic after this can be moved to the plugin manager

opensearch_single_kernel/events/opensearch.py
```diff
@@ -508,7 +530,11 @@ def _on_leader_elected(self, event: LeaderElectedEvent) -> None:  # noqa: C901
             return
 
         if not self.charm.state.application.is_admin_user_initialized:
-            self.charm.status.set(CharmStatuses.ADMIN_USER_INIT_IN_PROGRESS)
+            self.charm.state.add_status_if_not_present(
+                InternalUsersStatuses.ADMIN_USER_INIT_IN_PROGRESS.value,
```
akram09, line 535: Where do we clear this status ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -613,11 +659,14 @@ def _on_start(self, event: StartEvent) -> None:  # noqa: C901
...
         ):
-            self.charm.status.set(CharmStatuses.PEER_CLUSTER_NO_DATA_NODE)
+            self.charm.state.add_status_if_not_present(
+                PeerClusterStatuses.PEER_CLUSTER_NO_DATA_NODE.value,
+                "unit",
+                self.charm.cluster_manager.name,
+            )
             event.defer()
             return
         # We are requesting start of openSearch
```
akram09, line 670: This changes the UX no ?

opensearch_single_kernel/events/tls.py
```diff
@@ -151,14 +151,25 @@ def _on_tls_relation_created(self, event: RelationCreatedEvent) -> None:
         self.certs.request_certificate_creation(certificate_signing_request=unit_transport_csr)
         self.certs.request_certificate_creation(certificate_signing_request=unit_http_csr)
 
+        self.charm.state.remove_status_if_present(
```
akram09, line 154: This should be done at manager no ?

opensearch_single_kernel/events/tls.py
```diff
@@ -151,14 +151,25 @@ def _on_tls_relation_created(self, event: RelationCreatedEvent) -> None:
...
+            TlsStatuses.TLS_RELATION_MISSING.value, "unit", self.charm.tls_manager.name
+        )
+        if not self.charm.tls_manager.all_tls_resources_stored():
```
akram09, line 157: How can we have all tls resources stored whereas we just got relation created

opensearch_single_kernel/managers/cluster.py
```diff
@@ -721,3 +731,51 @@ def get_prometheus_labels(self) -> dict[str, str] | None:
...
+        if not self.state.server.started:
+            status_list.append(GeneralStatuses.WAITING_TO_START.value)
+
+    def _add_app_statuses(self, status_list: list[StatusObject]) -> None:
+        """Compute the manager's app statuses and append them to list."""
+        if (
+            deployment_desc := self.state.application.deployment_desc
+        ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            if self.state.oauth_relation:
+                status_list.append(GeneralStatuses.OAUTH_RELATION_INVALID.value)
```
akram09: Why do we have these here ?

opensearch_single_kernel/managers/cluster.py
```diff
@@ -721,3 +731,51 @@ def get_prometheus_labels(self) -> dict[str, str] | None:
...
+        self, scope: AdvancedStatusesScope, recompute: bool = False
+    ) -> list[StatusObject]:
+        """Compute the manager's statuses."""
+        if not recompute:
+            return self.state.statuses.get(scope, self.name).root or [
+                GeneralStatuses.ACTIVE_IDLE.value
+            ]
+
+        status_list: list[StatusObject] = []
+
```
akram09, line 744: I think we should also add checks here is security index initialised and if admin user is initialized

opensearch_single_kernel/managers/health.py
```diff
@@ -99,3 +106,106 @@ def wait_for_shards_relocation(self) -> None:
             # we throw an error because various operations should NOT start while data
             # is being relocated. Examples are: simple stop, unit removal, upgrade
             raise OpenSearchHAError("Shards haven't completed relocating.")
+
+    def apply_health(
```
akram09, line 110: This is nice

opensearch_single_kernel/managers/lock.py
```diff
@@ -66,8 +70,15 @@ def acquire(self) -> bool:
                 "[Node lock] Not acquired. Unit with peer databag lock: %s",
                 self.state.application_lock.unit_with_lock,
             )
+            self.state.add_status_if_not_present(
```
akram09: Why not add it to the LockManager instead ?

akram09 (commented):
Thank you Serhii, The PR seems in a good shape already. I have left my comments. 

I do think that we can further optimize the code in events handler now that we can add/remove status using the state without making the manager aware about the charm.

opensearch_single_kernel/charms/base.py
s-mlnk: No. You can check `utils/status.py` for the code origin.

opensearch_single_kernel/common/client.py
s-mlnk: There is in health manager:
```python
host = self.state.host_ip if use_localhost else None
response = self.opensearch_client.get_health(host, wait_for_green_first, self.alt_hosts)
```

opensearch_single_kernel/core/state.py
s-mlnk: I've moved them to the charm base and passed through initialiser to state.

opensearch_single_kernel/core/state.py
s-mlnk: 1. Parameter name was unified both across `add_status_if_not_present` and `remove_status_if_present`, and in latter case `override_parameters` wouldn't be a great fit.
2. We cannot put dictionary as a parameter default value.

opensearch_single_kernel/core/state.py
s-mlnk: You wouldn't do a search by parameters for a statuses with the single dynamic parameter, however you will need an interpolated removal. For example: `Example status error: {}`. On the other hand, it would be helpful to differentiate between statuses with multiple ones: `Example status error on relation {}: {}`.

opensearch_single_kernel/events/external_clients.py
s-mlnk: Running statuses are set specifically by `set_running_status` function and not the `statuses.add` that is used in state.

opensearch_single_kernel/events/jwt.py
s-mlnk: Removed.

opensearch_single_kernel/events/jwt.py
s-mlnk: OAuth didn't have the status setting, so I also removed it from JWT to unify them.

opensearch_single_kernel/events/notifications.py
s-mlnk: It depends on notification & keystore managers so I cannot do that.

opensearch_single_kernel/events/opensearch.py
s-mlnk: In the internal users manager.

opensearch_single_kernel/events/opensearch.py
s-mlnk: What exactly? Here I only migrated to advanced status lib.

opensearch_single_kernel/events/tls.py
s-mlnk: I don't think we have alternative function for relation creation in tls_manager to put it into. `create_certificate_signing_request` is the most fitting, but it also called on certificate expire, so not quite a full match.

opensearch_single_kernel/events/tls.py
s-mlnk: If it were re-relation. Opensearch logic is next: even if relation broken, we still maintain TLS resources and operate in active / idle status up to the moment of the certificate expiration.

opensearch_single_kernel/managers/cluster.py
s-mlnk: We don't have a JWT & OAuth specific managers, so I put them in "general" cluster manager.

opensearch_single_kernel/managers/cluster.py
s-mlnk: They both are running statuses and therefore don't need to be recomputed.

opensearch_single_kernel/managers/lock.py
s-mlnk: I made it running status so now it's set in events.

opensearch_single_kernel/charms/base.py
Mehdi-Bendriss: @akram09 Sure, that sounds sensible. i'm thinking whether cluster and profiles should be swapped? nonetheless your order makes sense and we can improve it in the future as we see how it goes

opensearch_single_kernel/charms/base.py
```diff
@@ -70,11 +82,15 @@ class OpenSearchBaseCharm(ops.CharmBase, ABC):
...
-
         # State
-        self.state = ClusterState(self, self.substrate)
+        self.state = ClusterState(
+            self,
+            self.substrate,
+            SmtpRequires(self, SMTP_RELATION),
+            S3Requirer(self, S3_RELATION),
+            AzureStorageRequires(self, AZURE_RELATION),
+            GcsStorageRequires(self, GCS_RELATION),
```
Mehdi-Bendriss, line 92: are these needed to be passed?

opensearch_single_kernel/common/statuses.py
```diff
@@ -8,147 +8,236 @@
...
-        "Too many object storage relations. Only one is supported."
+    PEER_CLUSTER_NO_RELATION = StatusObject(
+        status="blocked", message="Cannot start. Waiting for peer cluster relation..."
     )
-    BACKUP_RELATION_DATA_INCOMPLETE = BlockedStatus("Backup relation data missing or incomplete.")
-    BACKUP_CREDENTIALS_INCORRECT = BlockedStatus(
-        "Backup configuration error: bad credentials, permissions, invalid CA, or unsupported configuration."
+    PEER_CLUSTER_WRONG_RELATION = StatusObject(
+        status="blocked",
+        message="Cluster name don't match with related cluster. Remove relation.",
```
Mehdi-Bendriss:
```suggestion
        message="Cluster name doesn't match with related cluster. Remove relation.",
```

opensearch_single_kernel/core/state.py
```diff
@@ -144,6 +160,16 @@ def started(self) -> str:
         """Get the value of 'started' key from unit data bag"""
         return self.relation.data[self.unit].get("started", "")
 
+    @started.setter
+    def started(self, value: str) -> None:
+        """Get the value of 'started' key from unit data bag"""
+        self.relation_data.update({"started": value})
+
+    @started.deleter
+    def started(self) -> None:
```
Mehdi-Bendriss, line 169: missing docstrings

opensearch_single_kernel/core/state.py
```diff
@@ -144,6 +160,16 @@ def started(self) -> str:
         """Get the value of 'started' key from unit data bag"""
         return self.relation.data[self.unit].get("started", "")
 
+    @started.setter
+    def started(self, value: str) -> None:
+        """Get the value of 'started' key from unit data bag"""
```
Mehdi-Bendriss:
```suggestion
        """Set the value of the 'started' key."""
```

opensearch_single_kernel/core/state.py
```diff
@@ -787,24 +820,38 @@ def trigger_relation_changed(self) -> None:
...
         self.peer_app_interface = DataPeerData(model=charm.model, relation_name=PEER_RELATION)
         self.peer_unit_interface = DataPeerUnitData(model=charm.model, relation_name=PEER_RELATION)
         self.client_data_interface = OpenSearchProvidesData(
             model=charm.model, relation_name=CLIENT_RELATION
         )
 
+        self.smtp_requires = smtp_requires
+        self.s3_requirer = s3_requirer
+        self.azure_requires = azure_requires
+        self.gcs_requires = gcs_requires
```
Mehdi-Bendriss, line 854: @akram09 according to the code spec, should we pass these are constructor args, or should we instantiate them here or ..?

opensearch_single_kernel/core/state.py
```diff
@@ -1008,14 +1070,12 @@ def is_tls_full_configured_in_cluster(self) -> bool:
...
             return False
-        for unit in self.all_units:
-            if (
-                self.peer_relation.data[unit].get("tls_configured", "").lower() != "true"
-                or "tls_ca_renewing" in self.peer_relation.data[unit]
-                or "tls_ca_renewed" in self.peer_relation.data[unit]
-            ):
-                return False
-        return True
+        return all(
```
Mehdi-Bendriss, line 1074: nice

opensearch_single_kernel/events/jwt.py
```diff
@@ -140,11 +170,7 @@ def apply_security_config_if_needed(self, event: EventBase) -> None:
...
             # we need to come back in this case because there will not be a follow-up event
```
Mehdi-Bendriss, line 175: Could we add back the log line 
`logger.debug(f"Error when updating the security index: {e.out}")`

opensearch_single_kernel/events/notifications.py
Mehdi-Bendriss, line 297: can you check the hook `_on_secret_changed` down below? 
There is a part:
``` 
self.charm.plugin_manager.put_plugin_config(
            scope=Scope.UNIT,
            label=label,
            cleanup={
                "keys": list(keys.keys()),     # keys is resolved as a string, not a map. So `.keys()` is problematic
                "smtp_account_id": smtp_account_id_list,
            },
        )
```
can you double check the `keys` variable type?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -316,7 +325,7 @@ def _on_update_status(self, event: UpdateStatusEvent) -> None:  # noqa: C901
             logger.debug("Deployment description not yet computed")
             return
 
-        if self.check_profile_missing_requirements():
+        if not self.check_profile_requirements():
```
Mehdi-Bendriss, line 328: good catch

opensearch_single_kernel/events/opensearch.py
```diff
@@ -695,7 +758,7 @@ def _on_start_opensearch(self, event: StartOpenSearch) -> None:  # noqa: C901
         # - cluster health
         if not all(
             [
-                not self.check_profile_missing_requirements(),
+                self.check_profile_requirements(),
```
Mehdi-Bendriss, line 761: good catch!

opensearch_single_kernel/events/opensearch.py
```diff
@@ -895,49 +985,69 @@ def is_cluster_healthy_to_start(self) -> bool:
...
 
-        Requirements include:
-        - System requirements
-        - Memory requirements
-        - Cluster topology requirements
+        Returns:
+            whether the profile passed validation and all requirements.
         """
         try:
             self.charm.profiles_manager.config_profile
```
Mehdi-Bendriss, line 995: what does this do?

opensearch_single_kernel/events/snapshots.py
```diff
@@ -324,22 +367,26 @@ def _on_verify_snapshots_credentials(  # noqa C901
             logger.warning(
                 "Object storage configuration not ready for backup credentials verification."
             )
-            if self.charm.unit.is_leader():
-                self.charm.status.set(CharmStatuses.BACKUP_RELATION_DATA_INCOMPLETE, app=True)
+            self.charm.state.add_status_if_not_present(
```
Mehdi-Bendriss, line 370: question here, wouldn't the `get_statuses` of the manager not compute many of these states any way and report them? in which case, is there a reason to add them here as well?

opensearch_single_kernel/managers/tls.py
```diff
@@ -560,3 +566,35 @@ def get_secrets_for_cert_type(self, cert_type: CertType) -> dict[str, str]:
...
+                status_list.append(TlsStatuses.TLS_CA_ROTATION.value)
+            if not self.all_tls_resources_stored():
+                status_list.append(
+                    TlsStatuses.TLS_NOT_FULLY_CONFIGURED.value
+                    if self.state.tls_relation
+                    else TlsStatuses.TLS_RELATION_MISSING.value
+                )
+            if not self.state.tls_relation and (certs := self.check_certs_expiration()):
+                missing = [cert.val for cert in certs.keys()]
+                self.state.add_status_if_not_present(
```
Mehdi-Bendriss: self.state or `status_list.append`?

tests/integration/ha/test_backup.py
```diff
@@ -499,8 +500,6 @@ async def test_large_deployment_build_and_deploy(
     await wait_until(
         ops_test,
         apps=[TLS_CERTIFICATES_APP_NAME, "main", "failover", APP_NAME],
-        apps_statuses=["active"],
-        units_statuses=["active"],
```
Mehdi-Bendriss, line 503: wasn't this correct?

tests/integration/ha/test_backup.py
```diff
@@ -652,8 +645,6 @@ async def test_large_setups_relations_with_misconfiguration(  # noqa: C901
         await wait_until(
             ops_test,
             apps=["main"],
-            units_statuses=["active"],
-            apps_statuses=["active"],
```
Mehdi-Bendriss, line 656: same question, applies to all occurrences

tests/integration/ha/test_backup.py
```diff
@@ -1458,11 +1422,7 @@ async def test_wrong_azure_credentials(
     await wait_until(
         ops_test,
         apps=[app],
-        units_statuses=["active"],
```
Mehdi-Bendriss, line 1461: applies to all occurrences, if the app status is blocked, but the unit status is Active, we should keep as-is

tests/integration/profiles/test_profiles.py
```diff
@@ -159,20 +157,7 @@ async def test_insufficient_memory(ops_test: OpsTest, charm: str, series: str) -
     await wait_until(
         ops_test,
         apps=[APP_NAME],
-        apps_full_statuses={
```
Mehdi-Bendriss, line 162: why not test the app statuses as well as unit statuses?

Mehdi-Bendriss (commented):
Thanks Serhii! That's great work! 
I left some comments and questions on integration tests where we remove app/unit statuses in checks

opensearch_single_kernel/charms/base.py
s-mlnk: Yes. Check also https://github.com/canonical/opensearch-single-kernel-library/pull/55#discussion_r3092328961.

opensearch_single_kernel/core/state.py
akram09: I would advocate to instantiate them in charm and pass them to state.

s-mlnk: Check https://github.com/canonical/opensearch-single-kernel-library/pull/55#discussion_r3092328961.

opensearch_single_kernel/events/notifications.py
s-mlnk: While your assumptions are right for plugin content, it is different for the plugin secret: https://github.com/canonical/opensearch-single-kernel-library/blob/0a28c200e4afc1e0f00fffa4d62cda5393dd9293/opensearch_single_kernel/managers/plugin.py#L61-L73 It was done in the original codebase so I left it as-is.

opensearch_single_kernel/events/opensearch.py
s-mlnk: I mean I inverted this function 😅

opensearch_single_kernel/events/opensearch.py
s-mlnk: I mean I inverted this function 😅

tests/integration/ha/test_backup.py
s-mlnk: Because I reworked `wait_until` function and now it waits for active idle statuses by default.

tests/integration/ha/test_backup.py
s-mlnk: Because I reworked `wait_until` function and now it waits for active idle statuses by default.

tests/integration/ha/test_backup.py
s-mlnk: Because I reworked `wait_until` function and now it waits for active idle statuses by default.

opensearch_single_kernel/events/opensearch.py
s-mlnk: It builds the profile model from config and if it's invalid an exception will be thrown.

opensearch_single_kernel/events/snapshots.py
s-mlnk: Here the status is added immediately in the hook, so after the hook ends status is already set to the charm. Without that and until the next update-status event, charm will be in false-positive active idle status, that breaks UX and may break integration tests.

opensearch_single_kernel/managers/tls.py
s-mlnk: Nice catch!

tests/integration/profiles/test_profiles.py
s-mlnk: Because I reworked `wait_until` function and now it waits for active idle statuses by default.

opensearch_single_kernel/charms/base.py
s-mlnk: Done

Mehdi-Bendriss (dismissed):
Thanks Serhii! All good for me. 
One last request from me is about testing it with `status-detail recompute=true`, and seeing that the behavior is maintained.

akram09 (dismissed):
Seems good for me

akram09 (dismissed):

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -272,3 +285,66 @@ def version(self) -> str:
...
+            if extra_user_roles != KIBANA_SERVER_ROLE and not self.opensearch_client.get_user(
+                external_client.relation_username
+            ):
+                status_list.append(
+                    format_status(
+                        ExternalClientsStatuses.USER_CREATION_FAILED.value, {"id": relation.id}
+                    )
+                )
+                return
+        except OpenSearchHttpError:
```
Mehdi-Bendriss: please just add an error log and we're good

Mehdi-Bendriss (commented):
Good catch Serhii! one minor final comment from me

Mehdi-Bendriss (approved):

PR #58, opened 2026-04-09
patch: [DPE-9264] Integrate Data Interfaces v1

Pepsiqqq:
Implemented data interface v1 
Changed the way we interact with databag to models, so now you don't need to create setters/getters for new fields/secrets.

Split state into state and cluster, logic same just two smaller files instead of one

Removed secret management, all of this is now done by data_interface library

Refactored external clients to use new classes/methods, also moved everything related to external clients from events/opensearch.py to events/external_clients.py

Refactored JWT to also use models and added support to both data_interfaces V0 and V1 providers(do v1 providers exist?)

Small changes to plugins secrets:
1. Now plugin secret is a string which can be json.loads into dict [str,str].
2. First string is plugin name(I don't know if there can be two plugins with same name, right now we only have SMTP and when SMTP creates plugin name it's also adds relation-id to it) and second is secret content
3. The PluginConfigInfo have secret_id changed to secret_name, by which we lookup secret. 

Refactored TLS:
We no longer can parse all the secret into one dict and just take stuff from it, now we need to specify which cert_type(admin, http, transport) and field(key, cls, cert, etc.) do we need.
 
Updated integration test for new secrets label
And also add application with data_interfaces v1 to test in opensearch_provider.py how OpenSearch works with both v1 and v0 databags

Deleted secrets unit test, as we no longer have secret management, instead add small test in test_charm to see if secrets are written to databag


opensearch_single_kernel/common/client.py
```diff
@@ -34,7 +34,11 @@
     Substrates,
 )
 from opensearch_single_kernel.common.exceptions import OpenSearchHttpError
-from opensearch_single_kernel.core.models import App, Node, ObjectStorageConfig
+from opensearch_single_kernel.core.plain_base import (
```
akram09: nit: I would rename plain_base to base_models or just models

opensearch_single_kernel/common/constants.py
```diff
@@ -209,6 +209,14 @@ class ExtraUserRolePermissions(Enum):
 OPENSEARCH_SYSTEM_USERS = {ADMIN_USER, KIBANA_SERVER_USER}
 OPENSEARCH_USERS = OPENSEARCH_SYSTEM_USERS | {COS_USER}
 
+# Maps an internal user to its (password, hashed_password) field names on
+# OpenSearchAppPeerUserSecretsModel.
+USER_SECRET_FIELDS: dict[str, tuple[str, str]] = {
```
akram09, line 214: I don't remember we had a cos_password ? Or is it a typo and you meant monitor ?

opensearch_single_kernel/charms/base.py
```diff
@@ -51,6 +52,10 @@
 from opensearch_single_kernel.events.snapshots import SnapshotsEventsHandler
 from opensearch_single_kernel.events.tls import TLSEventsHandler
 from opensearch_single_kernel.events.upgrades import UpgradesEventsHandler
+from opensearch_single_kernel.lib.charms.data_platform_libs.v1.data_interfaces import (
```
akram09: I think we should use charmlibs library no ?

opensearch_single_kernel/core/jwt.py
```diff
@@ -0,0 +1,25 @@
...
+)
+
+
+class JWTAuthConfiguration(ResourceProviderModel):
+    """Model class for the configuration parameters of JWT authentication."""
+
+    signing_key: ExtraSecretStr = Field(default=None)
+    jwt_header: str | None = Field(default=None)
+    jwt_url_parameter: str | None = Field(default=None)
+    roles_key: str | None = Field(default=None)
```
akram09: I think this was non nullable

opensearch_single_kernel/core/jwt.py
```diff
@@ -0,0 +1,25 @@
...
+
+from pydantic import Field
+
+from opensearch_single_kernel.lib.charms.data_platform_libs.v1.data_interfaces import (
+    ExtraSecretStr,
+    ResourceProviderModel,
+)
+
+
+class JWTAuthConfiguration(ResourceProviderModel):
```
akram09, line 14: Why should we use ResourceProviderModel ?

opensearch_single_kernel/core/relation_base.py
```diff
@@ -0,0 +1,378 @@
...
+    OptionalSecretStr, Field(exclude=True, default=None), SECRET_UNIT_TRANSPORT
+]
+HttpSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_UNIT_HTTP]
+AdminSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_APP_ADMIN]
+UserSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_USERS]
+PluginsSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_PLUGIN]
+BackupSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_BACKUPS]
+
+
+class RelationModel(BaseModel):
```
akram09, line 45: What is the problem we are trying to solve with this ?

opensearch_single_kernel/core/lock.py
```diff
@@ -0,0 +1,67 @@
...
+class LockAppStateModel(RelationModel, PeerModel):
+    """Peer model mapping to the Lock application state."""
+
+    # Juju event id during which the leader granted the lock to itself; the leader may
+    # only use the lock in a next event (see grant_lock).
+    leader_acquired_lock_after_juju_event_id: str | None = Field(default=None)
+    # Name of the unit currently holding the peer lock, None when the lock is free.
+    unit_with_lock: str | None = Field(default=None)
+
+    def grant_lock(self, unit_name: str, own_unit_name: str) -> None:
```
akram09, line 26: Why do we have this implemented here? This should be in the state I think

opensearch_single_kernel/core/lock.py
```diff
@@ -0,0 +1,67 @@
...
+
+class LockServerStateModel(RelationModel, PeerModel):
+    """Peer model mapping to the Lock unit state."""
+
+    # Whether this unit is asking the leader for the peer lock.
+    lock_requested: bool = Field(default=False)
+    # Write-only field to force a peer relation-changed event on other units.
+    trigger: str | None = Field(default=None, alias="-trigger")
+
+    def trigger_relation_changed(self) -> None:
```
akram09, line 63: Same for this I think the model should not be aware of this detail

opensearch_single_kernel/core/relation_base.py
```diff
@@ -0,0 +1,378 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+TransportSecretStr = Annotated[
+    OptionalSecretStr, Field(exclude=True, default=None), SECRET_UNIT_TRANSPORT
+]
+HttpSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_UNIT_HTTP]
+AdminSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_APP_ADMIN]
+UserSecretStr = Annotated[OptionalSecretStr, Field(exclude=True, default=None), SECRET_USERS]
```
akram09: If I am not wrong I think this is already implemented in data-interfaces v1

opensearch_single_kernel/core/peer_app.py
```diff
@@ -0,0 +1,268 @@
...
+    nodes_config: dict[str, Node] = Field(default_factory=dict, alias="nodes_config")
+    # Whether the application-level cluster bootstrap process has completed.
+    bootstrapped: bool = Field(default=False)
+    # Description of application's role/config within the deployment.
+    deployment_description: DeploymentDescription | None = Field(default=None)
+    # Peer-cluster fleet apps discovered by this application, keyed by app id.
+    cluster_fleet_apps: dict[str, PeerClusterApp] = Field(
+        default_factory=dict, alias="cluster_fleet_apps"
+    )
+    # Peer-cluster fleet apps learned through peer-cluster relations (from other apps in the
```
akram09, line 87: This is the peer cluster rels it is keyed by relation id

opensearch_single_kernel/core/peer_app.py
```diff
@@ -0,0 +1,268 @@
...
+            if not plugin_m.plugin_secrets:
+                plugin_m.plugin_secrets = "{}"
+        if user_m := self.build_sibling_model(OpenSearchAppPeerUserSecretsModel):
+            if not user_m.admin_password:
+                user_m.admin_password = " "
+        if admin_tls_m := self.build_sibling_model(OpenSearchAppPeerAdminTlsSecretsModel):
+            if not admin_tls_m.admin_key_password:
+                admin_tls_m.admin_key_password = " "
+
+    def get_user_secret(self, user: str, hashed: bool = False) -> str | None:
```
akram09, line 172: We don't really need to replicate this, It was a simplification before. We can rely now directly on getting the user we want with fields

opensearch_single_kernel/core/peer_app.py
```diff
@@ -0,0 +1,268 @@
...
+        field_name = fields[1] if hashed else fields[0]
+        user_m = self.build_sibling_model(OpenSearchAppPeerUserSecretsModel)
+        value = getattr(user_m, field_name) if user_m else None
+        # admin_password may hold a single-space placeholder to force secret
+        # creation (see initialize_empty_secrets)
+        if user == ADMIN_USER and not hashed:
+            return stripped_or_none(value)
+        return value
+
+    def to_peer_cluster_rel_data(
```
akram09, line 186: Can you remind me why we need this ?

opensearch_single_kernel/core/peer_app.py
```diff
@@ -0,0 +1,268 @@
...
+                    or ""
+                )
+                if is_main_orchestrator
+                else ""
+            ),
+        }
+
+        return PeerClusterAppModel(**copied_data)
+
+    def update_from_peer_cluster_rel_data(self, peer_data: PeerClusterAppModel) -> None:
```
akram09, line 223: I don't know if it is worth delegating such logic to model. Do you have arguments ?

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,350 @@
...
+)
+from opensearch_single_kernel.lib.charms.data_platform_libs.v1.data_interfaces import (
+    BaseCommonModel,
+    PeerModel,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PeerClusterRelErrorData(PlainModel):
```
akram09, line 47: Why use the PlainModel ?

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,350 @@
...
+    # Hash of the last snapshots (backup) credentials this unit persisted to its keystore.
+    snapshots_credentials_saved: Optional[str] = Field(default=None)
+
+    @property
+    def unit(self):
+        """The ops.Unit this model is bound to."""
+        return self.component
+
+
+class PeerClusterAppModel(RelationModel, BaseCommonModel):
```
akram09, line 139: Why not PeerAppModel ?

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,350 @@
...
+    cross-cluster sharing.
+    """
+
+    # Whether the requirer app offers itself as a failover-orchestrator candidate.
+    is_candidate_failover_orchestrator: bool = Field(default=False)
+    # Which orchestrator role ("main"/"failover") this relation was established for.
+    trigger: Optional[str] = Field(default=None)
+    # Whether the requirer has acknowledged/registered the main orchestrator.
+    main_orchestrator_registered: Optional[bool] = Field(default=None)
+    # The current app's own identity/roles/unit-count on this relation.
```
akram09, line 153: The current peer cluster app

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,350 @@
...
+    cluster_fleet_apps: dict[str, PeerClusterApp] = Field(default_factory=dict)
+    # The currently elected main/failover orchestrator pair. None means this relation has
+    # never had orchestrator data written to it.
+    orchestrators: Optional[PeerClusterOrchestrators] = Field(default=None)
+    # Hash of the last broadcast payload
+    rel_data_hash: Optional[str] = Field(default=None)
+    error_data: Optional[PeerClusterRelErrorData] = Field(default=None)
+    security_index_initialised: bool = Field(default=False)
+    first_data_node: Optional[str] = Field(default=None)
+    cluster_name: str = Field(default="")
```
akram09: We don't have cluster_name in Peer cluster app databag

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,350 @@
...
+    nodes_config: dict[str, Node] = Field(default_factory=dict)
```
akram09, line 165: These fields used to be in PeerclusterData

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,350 @@
...
+            m.s3_secret_key = None
+            m.s3_tls_ca_chain = None
+            m.azure_storage_account = None
+            m.azure_secret_key = None
+            m.gcs_secret_key = None
+            # emptying the secret fields above deletes the backing group secrets, so
+            # drop the marker to let initialize_empty_secrets() re-create them later
+            m.pc_secrets_initialized = False
+
+    def initialize_empty_secrets(self) -> None:
```
akram09, line 327: Is this only for the log ? If that is the case is it an error log ?

opensearch_single_kernel/core/peer_secrets.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class OpenSearchAppPeerUserSecretsModel(RelationModel, PeerModel):
+    """Peer model mapping to the OpenSearch application's internal-user credentials.
+
+    Split out of OpenSearchAppPeerModel so that reading/writing a plain
+    application field doesn't also resolve the Juju secret group.
+    """
+
+    admin_password: UserSecretStr = Field(default="")
```
akram09: Can't we include these fields directly in the Model using it for instance the AppPeerModel ?

opensearch_single_kernel/core/peer_secrets.py
```diff
@@ -0,0 +1,105 @@
...
+    admin_hashed_password: UserSecretStr = Field(default="")
```
akram09: Is this a breaking change, Since I remember we were having a secret for each field

opensearch_single_kernel/core/peer_unit.py
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    # Aliases here are pinned to the underscored keys deployed databags use,
+    # so upgrade works correctly
+
+    # Performance profile ("testing"/"production") applied to this unit's JVM/OpenSearch config.
+    # None means "not yet set" callers fall back to the profile configured via charm config.
+    profile: Optional[PerformanceType] = Field(default=None)
+    # Whether this unit was one of the initial seed nodes used to bootstrap the cluster.
+    bootstrap_contributor: bool = Field(default=False, alias="bootstrap_contributor")
```
akram09, line 70: nit: We can keep the name as it was ? `is_bootstrap_contributor` and use the alias

opensearch_single_kernel/core/peer_unit.py
```diff
@@ -0,0 +1,162 @@
...
+    # Whether this unit is currently mid CA-rotation
+    tls_ca_renewing: bool = Field(default=False, alias="tls_ca_renewing")
+    # Whether this unit has finished renewing to the new CA.
+    tls_ca_renewed: bool = Field(default=False, alias="tls_ca_renewed")
+    # Whether this unit's TLS certificates are fully configured.
+    tls_configured: bool = Field(default=False, alias="tls_configured")
+    # Last time application's databag was updated; used to force relation-changed hook
+    update_ts: str = Field(default="")
+    # Timestamp of the last time this unit checked its certificates for upcoming expiry.
+    certs_exp_checked_at: str = Field(default="1970-01-01 00:00:00", alias="certs_exp_checked_at")
```
akram09, line 85: Why the default value is set to that ?

opensearch_single_kernel/core/peer_unit.py
```diff
@@ -0,0 +1,162 @@
...
+    plugin_config_info: dict[str, PluginConfigInfo] = Field(
+        default_factory=dict, alias="plugin_config_info"
+    )
+    oauth_openid_connect_url: str = Field(default="", alias="oauth_openid_connect_url")
+    # Set when this specific unit is departing/scaling down. Used to skip relation-broken
+    # triggered by the unit's own removal.
+    unit_dying: bool = Field(default=False)
+    # PID of this unit's running pebble-observer subprocess, or None if not started/stopped.
+    pebble_observer_pid: Optional[int] = Field(default=None)
+
```
akram09, line 106: There was jwt_auth_configuration in the unit databag

opensearch_single_kernel/core/peer_unit.py
```diff
@@ -0,0 +1,162 @@
...
+    allocation_exclusions_to_delete: set[str] = Field(default_factory=set)
+    # Voting-exclusion entries application still needs to remove from the cluster voting config.
+    delete_voting_exclusions: set[str] = Field(default_factory=set)
+    # Last known IP address of this unit.
+    last_host_ip: str = Field(default="", alias="last_host_ip")
+    # Plugin configuration metadata unit is responsible for, key is plugin label
+    plugin_config_info: dict[str, PluginConfigInfo] = Field(
+        default_factory=dict, alias="plugin_config_info"
+    )
+    oauth_openid_connect_url: str = Field(default="", alias="oauth_openid_connect_url")
```
akram09, line 97: There is also "oauth_departing"

opensearch_single_kernel/core/peer_unit.py
```diff
@@ -0,0 +1,162 @@
...
+            ProductionProfile() if self.profile == PerformanceType.PRODUCTION else TestingProfile()
+        )
+
+    @opensearch_profile.setter
+    def opensearch_profile(self, value: OpenSearchProfile) -> None:
+        """Set current profile of the unit from an OpenSearchProfile instance."""
+        self.profile = value.type
+
+    @property
+    def is_app_leader(self) -> bool:
```
akram09: Make sense to put this directly in the state

opensearch_single_kernel/core/relation_base.py
```diff
@@ -0,0 +1,378 @@
...
+       the databag. You never call _write_to_databag() yourself.
+
+           server.started = "1232323" -> databag: started = 1232323
+
+       Some writes are skipped and kept in-memory only (see `_writable_repository`): an unbound
+       model, a read-only one (points at another app's databag we can't write), or an app
+       databag when we're not the leader.
+
+    3) SECRET FIELDS is delegated to a "sibling" models
+       Some fields are not plain databag values because they belong to a Juju secret group. To
```
akram09: Sorry for my lack of context, What does a juju secret group means ?

opensearch_single_kernel/core/relation_base.py
```diff
@@ -0,0 +1,378 @@
...
+        """
+        self._update_depth_counter += 1
+        try:
+            yield self
+        finally:
+            self._update_depth_counter -= 1
+            if self._update_depth_counter == 0:
+                self._write_to_databag()
+
+    @contextmanager
```
akram09: This seems deprecated

opensearch_single_kernel/core/state.py
```diff
@@ -214,10 +223,54 @@ def upgrade_relation(self) -> Relation | None:
...
+    def _model_key(model_cls: type, relation: Relation | None, component) -> tuple:
+        """Cache key identifying the logical databag a model maps to."""
+        return (
+            model_cls,
+            relation.name if relation else None,
+            relation.id if relation else None,
+            getattr(component, "name", None),
+        )
+
+    def _cached_model(self, key: tuple, factory):
```
akram09: Why we need this ?

opensearch_single_kernel/core/state.py
```diff
@@ -684,23 +759,12 @@ def ca_and_certs_rotation_complete_in_cluster(self) -> bool:
...
-        return all(
-            [
-                server.tls_configured and (not server.tls_ca_renewing or server.tls_ca_renewed)
-                for server in all_units_in_fleet
-            ]
+        complete = all(
+            server.tls_configured and (not server.tls_ca_renewing or server.tls_ca_renewed)
+            for server in all_units_in_fleet
         )
+        logger.debug("CA and certs rotation complete in cluster: %s", complete)
```
akram09, line 738: Can you add a debug for all the values. Can be very useful for debugging TLS issues

opensearch_single_kernel/core/state.py
```diff
@@ -993,8 +1049,8 @@ def is_peer_cluster_provider(self, typ: Literal["main", "failover"] | None = Non
             return False
 
         # check if the current app is elected orchestrator
-        if not (orchestrators := self.application.orchestrators):
-            # not populated yet
+        orchestrators = self.application.orchestrators
```
akram09, line 1024: Walrus

opensearch_single_kernel/events/external_clients.py
```diff
@@ -76,66 +81,118 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
         if not self.charm.unit.is_leader():
             return
 
-        if not self.charm.cluster_manager.opensearch_client.is_node_up() or not event.index:
-            event.defer()
-            return
-        if not (external_client := self.charm.state.external_client_by_relation(event.relation)):
-            logger.error("No external client found for relation id %d", event.relation.id)
+        index_name = self._validate_request(event)
+        if not index_name:
```
akram09: walrus

opensearch_single_kernel/events/external_clients.py
```diff
@@ -76,66 +81,118 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
-            self.charm.state.add_status_if_not_present(
-                ExternalClientsStatuses.INVALID_INDEX_NAME.value,
-                "unit",
-                self.charm.external_clients_manager.name,
-                dynamic_params={"id": event.relation.id, "index": event.index},
-            )
-            return
-        self.charm.state.remove_status_if_present(
-            ExternalClientsStatuses.INVALID_INDEX_NAME.value,
+        self.charm.status_handler.set_running_status(
```
akram09: This will be cleared once you rebase on status PR

opensearch_single_kernel/events/external_clients.py
```diff
@@ -146,67 +203,73 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
         self.charm.state.remove_status_if_present(
             ExternalClientsStatuses.USER_CREATION_FAILED.value,
             "unit",
             self.charm.external_clients_manager.name,
             interpolated=True,
             search_parameters={"id": event.relation.id},
         )
+        return credentials
+
+    def _get_connection_data(self, event: ResourceRequestedEvent) -> dict | None:
```
akram09: Can we move this to the Manager instead ?

opensearch_single_kernel/events/keystore.py
```diff
@@ -34,6 +35,16 @@ def __init__(
 
     def _on_reload(self, event: ReloadKeystoreEvent) -> None:
         """Handle keystore reload event."""
-        if not self.charm.keystore_manager.reload():
-            logger.error("Keystore reload failed.")
-            event.defer()
+        result = self.charm.keystore_manager.reload()
```
akram09, line 38: Why not return a boolean and raise exception if error ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -156,15 +156,15 @@ def _on_peer_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+        if not self.charm.state.application.deployment_description:
             logger.debug("Deployment description not yet computed.")
             return
 
-        if (
-            is_node_up := self.charm.cluster_manager.opensearch_client.is_node_up()
-            and self.charm.apply_health(app=self.charm.unit.is_leader())
-            in [HealthColors.UNKNOWN, HealthColors.YELLOW_TEMP]
-        ):
+        is_node_up = self.charm.cluster_manager.opensearch_client.is_node_up()
```
akram09, line 161: Why removed walrus ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -204,8 +200,7 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:  # noq
                 self.charm.peer_cluster_events.check_credentials_with_missing_relations()
                 if self.charm.state.peer_cluster_relations:
                     self.charm.peer_cluster_events.apply_orchestrator_status()
-
-        elif event.relation.data.get(event.app):
+        elif self.charm.state.application.nodes_config:
```
akram09: This is a bit different we were checking if the relation data has any data in it not for nodes_config

opensearch_single_kernel/events/opensearch.py
```diff
@@ -559,17 +540,20 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:  # noqa: C901
 
     def _on_leader_elected(self, event: LeaderElectedEvent) -> None:  # noqa: C901
         """Handle leader election event."""
+        if self.charm.unit.is_leader():
```
akram09, line 533: Can you explain this

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1314,45 +1297,69 @@ def _on_secret_changed(self, event: SecretChangedEvent) -> None:  # noqa: C901
...
-        # Variables for better readability
-        label_key = label_parts["key"]
-        is_leader = self.charm.unit.is_leader()
 
-        # Matching secrets by label
         if (
             label_parts["application_name"] != self.charm.app.name
             or label_parts["scope"] != Scope.APP
-            or label_key not in keys_to_process
+            or "admin-hashed-password" not in event.secret.get_content().keys()
```
akram09: Why not use constants like we were using ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1314,45 +1297,69 @@ def _on_secret_changed(self, event: SecretChangedEvent) -> None:  # noqa: C901
...
+            or "kibana-server-hashed-password" not in event.secret.get_content().keys()
         ):
             logger.info("Secret %s was not relevant for us.", event.secret.label)
             return
 
-        logger.debug("Secret change for %s", str(label_key))
+        logger.debug("Secret change for user secrets, updating all users.")
 
-        if is_leader and label_key == password_key(KIBANA_SERVER_USER):
+        if self.charm.unit.is_leader():
```
akram09: This can be moved to the external clients at this point

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1314,45 +1297,69 @@ def _on_secret_changed(self, event: SecretChangedEvent) -> None:  # noqa: C901
...
+        ):
+            return
+
+        try:
+            event.secret.get_info()
+        except SecretNotFoundError:
+            logger.info("Secret-remove for %s ignored, we are not the owner.", event.secret.label)
+            return
+
+        logger.debug("Removing obsolete revision of secret %s.", event.secret.label)
```
akram09, line 1278: Can you explain why ?

opensearch_single_kernel/events/peer_cluster.py
```diff
@@ -166,7 +162,12 @@ def _on_peer_cluster_orchestrator_relation_changed(  # noqa: C901
             self.charm.state.application.first_data_node = first_data_node
 
         # fetch emitting app planned units and broadcast
-        related_peer_cluster_app = PeerClusterApp.from_str(data.get("app"))
+        remote_peer_cluster = self.charm.state.peer_cluster_by_relation_id(
+            is_provider=True, relation_id=event.relation.id, remote=True
+        )
+        related_peer_cluster_app = remote_peer_cluster.app if remote_peer_cluster else None
+        if not related_peer_cluster_app:
```
akram09, line 164: Why are you returning here ?

opensearch_single_kernel/events/peer_cluster.py
```diff
@@ -365,24 +365,22 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
 
         if reconcile_deployment_desc:
             # check if valid data is present if so update the seed hosts
-            if data.get("data"):
+            if (
+                remote_peer_cluster is not None
+                and remote_peer_cluster.deployment_description is not None
```
akram09, line 363: Why do we check deployment_desc of the remote peer cluster ?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -237,11 +243,147 @@ def _on_upgrade_charm(self, event: UpgradeCharmEvent) -> None:
             get_charm_revision(self.charm.model.unit)
         )
         # TODO check backwards compatibility for profiles
+        self.charm.state.server.initialize_empty_secrets()
```
akram09, line 214: This should however only if we are upgrading from v0 instance to v1. What about rollbacks how can this be handled ?

opensearch_single_kernel/managers/cluster.py
```diff
@@ -946,7 +946,7 @@ def _add_app_statuses(self, status_list: list[StatusObject]) -> None:  # noqa: C
                 status_list.append(JwtStatuses.JWT_RELATION_INVALID.value)
         elif self.state.jwt_relation:
             try:
-                self.state.jwt.auth_configuration
+                _ = self.state.jwt
```
akram09: This is a bit weird, although I understand what you are trying to do

akram09 (commented):
Thank you Danylo, Some comments from me. I would need to go through the PR again after we have a call. I also didn't do manual testing yet

opensearch_single_kernel/core/jwt.py
Pepsiqqq: ResourceProviderModel is the data-interfaces v1 base for models read from a provider's databag

opensearch_single_kernel/core/relation_base.py
Pepsiqqq: It's the self-updating model base, assigning any field auto-serializes the whole model back
to its relation databag. This removes the need to hand-write a getter/setter per databag field. We also can add `update` for multiple fields

opensearch_single_kernel/core/lock.py
Pepsiqqq: grant_lock/release_lock are basically mutations of the lock databag fields (unit_with_lock, leader_acquired_lock_after_juju_event_id), I think if they only change owned databag they can stay with model

opensearch_single_kernel/core/peer_app.py
Pepsiqqq: no, `_on_get_password_action`, `put_or_update_internal_user_leader` is still dynamically gets user name. If we remove it we will need to add something like `if user == ADMIN_USER … elif …`  to these functions

opensearch_single_kernel/core/peer_app.py
Pepsiqqq: It marshals the local app peer model into a PeerClusterAppModel the data the orchestrator
broadcasts over the peer-cluster relation to the other clusters. It changed with di v1 because now we transfer secrets using lib without constructing json data blob or secret url ourselves

opensearch_single_kernel/core/peer_app.py
Pepsiqqq: From my point of view we only transferring data between clusters, changes to data still occur in events/managers so it should stay in state.

opensearch_single_kernel/core/peer_cluster.py
Pepsiqqq: it's old Model class just renamed

opensearch_single_kernel/core/peer_cluster.py
Pepsiqqq: The peer-cluster relation is a cross-application relation, so secret fields
must be stored as Juju secret URIs in the databag and granted to the remote app. the peer (PeerAppModel) base is for same-app peer databags and can't grant secrets cross app

opensearch_single_kernel/core/peer_cluster.py
Pepsiqqq: it's now consolidated into single model

opensearch_single_kernel/core/peer_cluster.py
Pepsiqqq: yes only for logs, no it's not an error log. But it makes debug logs unreadable

opensearch_single_kernel/core/peer_unit.py
Pepsiqqq: I think we should move away from old aliases to stay consistent with databag keys, now it's a lot easier because we can add value to model and it will be with same name in databag

opensearch_single_kernel/core/peer_unit.py
Pepsiqqq: this default value can be any number what is 6 h before opensearch deployment. I set it to unix epoch

opensearch_single_kernel/core/peer_unit.py
Pepsiqqq: The JWT auth config is now modelled by JWTAuthConfiguration (core/jwt.py), read from the JWT
provider relation rather than kept as a unit-databag field. It should work fine after upgrade because we will just read data again

opensearch_single_kernel/events/keystore.py
Pepsiqqq: Just for better readability, I can change it back

opensearch_single_kernel/events/opensearch.py
Pepsiqqq: Because is_node_up is used in more than one place. the health check on the same line and again
passed into get_nodes

opensearch_single_kernel/events/opensearch.py
Pepsiqqq: On leader election we create the app-level (and unit-level) secret groups by writing a
placeholder value into each group. Just to disable log spam from lib for missing secret groups

opensearch_single_kernel/events/opensearch.py
Pepsiqqq: v1 introduced library-side revision pruning that v0 never had, but this pruning excludes charm peer secrets so we do it ourselves

opensearch_single_kernel/events/peer_cluster.py
Pepsiqqq: the relation is gone/departing (remote_peer_cluster is None) or the remote hasn't written its app yet. Otherwise it will crash later. Essentially a None checker

opensearch_single_kernel/events/upgrades.py
Pepsiqqq: Rollback works fine but new secrets will be there. I don't know how we can clean them

opensearch_single_kernel/events/peer_cluster.py
Pepsiqqq: It distinguishes "provider sent an error but the relation still carries valid data" from "no
data at all". if the main orchestrator was scaled down to 0 and back, the error is
set but a valid deployment_description remains in the relation, in that case we still update the
seed hosts from the relation so data nodes pick up the new IPs instead of just severing.

opensearch_single_kernel/core/state.py
```diff
@@ -116,28 +106,48 @@ def __init__(
...
+            model=charm.model, relation_name=NODE_LOCK_RELATION, data_model=LockAppStateModel
         )
+        self.lock_unit_interface = OpsPeerUnitRepositoryInterface(
+            model=charm.model, relation_name=NODE_LOCK_RELATION, data_model=LockServerStateModel
+        )
+        self.jwt_interface = OpsRelationRepositoryInterface(
+            model=charm.model, relation_name=JWT_CONFIG_RELATION, data_model=JWTAuthConfiguration
+        )
+
+        self.opensearch_provides = opensearch_provider
```
skourta: General note. On other charms we usually do not have these in the state. We declare them on the event handlers.

opensearch_single_kernel/common/client.py
```diff
@@ -563,7 +567,13 @@ def remove_user_role(self, role_name: str) -> dict[str, Any]:
             HTTP response to opensearch API request.
         """
         try:
-            resp = self.request("DELETE", f"{USER_ROLE_ENDPOINT}/{role_name}")
+            resp = self.request(
+                "DELETE",
+                f"{USER_ROLE_ENDPOINT}/{role_name}",
+                retries=3,
+                wait_strategy=wait_fixed(3),
```
skourta: Can you explain why we need retry?

opensearch_single_kernel/common/constants.py
```diff
@@ -196,9 +196,11 @@ class ExtraUserRolePermissions(Enum):
 MAX_HEAP_SIZE_IN_KB = 31 * _1GB_IN_KB  # 31GB in KB
 PERFORMANCE_PROFILE = "profile"
 # Opensearch Snap revision
-OPENSEARCH_SNAP_REVISION = "98"  # Keep in sync with `workload_version` file
+# OPENSEARCH_SNAP_REVISION = "98"  # Keep in sync with `workload_version` file
 # Revision 2.19.5
 # OPENSEARCH_SNAP_REVISION = "108"  # Keep in sync with `workload_version` file
+# Revision 3.7.0
+OPENSEARCH_SNAP_REVISION = "49"  # Keep in sync with `workload_version` file
```
skourta: Are we changing snaps?
Why are reverting to 49? or using a new snap?

skourta: Version 3.7.0 is on revision 205 from `snap info opensearch` and 61 on `snap info opensearch-charmed`

skourta: Is the charm compatible with OS 3?

opensearch_single_kernel/core/lock.py
skourta: I do not see the value in this class. The lock is 2 fields that can be part of the app state. The functions should be in a lock manager instead.

opensearch_single_kernel/core/base_models.py
```diff
@@ -0,0 +1,233 @@
...
+            if isinstance(attr_val, list):
+                equal = equal and sorted(attr_val) == sorted(other_attr_val)
+            else:
+                equal = equal and (attr_val == other_attr_val)
+
+        return equal
+
+
+class App(PlainModel):
+    """Data class representing an application."""
```
skourta: This is not a data class but a pydantic model

opensearch_single_kernel/core/peer_app.py
skourta: +1 it should be `dict[int, PeerClusterApp]`

opensearch_single_kernel/core/peer_app.py
skourta: I would rather have a match statement. It is much cleaner and clearer

opensearch_single_kernel/core/peer_app.py
```diff
@@ -0,0 +1,242 @@
...
+            app for app in self.cluster_fleet_apps.values() if "data" in app.roles
+        ]
+        return bool(data_apps_in_fleet) and any(
+            app.planned_units > 0 for app in data_apps_in_fleet
+        )
+
+    def initialize_empty_secrets(self) -> None:
+        """Initialize empty app-level secrets to prevent log spam.
+
+        The v1 lib only creates a Juju secret when the written value is truthy.
```
skourta: How big of an impact does this optimisation have?

opensearch_single_kernel/core/peer_app.py
skourta: I think this should be a manager function

opensearch_single_kernel/core/peer_app.py
skourta: This also should be a manager function in my opinion.

opensearch_single_kernel/core/peer_cluster.py
skourta: I think we should separate the interfaces models from the "just" data models. This is confusing to read.

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,343 @@
...
+        return _sort_nested_dicts(value)
+
+    @model_serializer(mode="wrap")
+    def serialize_model(self, handler, info):
+        """Serialize the model, skipping secret resolution on request."""
+        if (info.context or {}).get("skip_secrets"):
+            return handler(self)
+        return BaseCommonModel.serialize_model(self, handler, info)
+
+    def apply_rel_data(self, source: "PeerClusterAppModel") -> None:
```
skourta: I do not get these functions

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,343 @@
...
+            m.admin_key_password = stripped_or_none(source.admin_key_password)
+            m.admin_csr = stripped_or_none(source.admin_csr)
+            m.admin_chain = stripped_or_none(source.admin_chain)
+            m.admin_cert = stripped_or_none(source.admin_cert)
+            m.admin_ca_cert = stripped_or_none(source.admin_ca_cert)
+
+            digest_source = source.model_dump(mode="json", context={"skip_secrets": True})
+            m.rel_data_hash = sha1(json.dumps(digest_source, sort_keys=True).encode()).hexdigest()
+
+    def set_backup_secrets(
```
skourta: Why isn't this a manager function?

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,343 @@
...
+                storage_account=self.azure_storage_account,
+                secret_key=self.azure_secret_key,
+            )
+        if cloud == "gcs":
+            if not self.gcs_secret_key:
+                return None
+            return GcsRelData.model_construct(secret_key=self.gcs_secret_key)
+        return None
+
+    def clear_rel_data(self) -> None:
```
skourta: Same. This should be a manager function

opensearch_single_kernel/core/relation_base.py
skourta: This feels like we are trying to rewrite data interfaces v1.
I think this is overengineered and not needed.
Something like [valkey or etcd](https://github.com/canonical/valkey-operator/blob/1a8f361456cec2fed2bf03fe90b27626fa636cbc/src/core/models.py#L374) is more than enough for most cases.

opensearch_single_kernel/core/state.py
```diff
@@ -214,65 +224,70 @@ def upgrade_relation(self) -> Relation | None:
...
-    @property
-    def loki_relation(self) -> Relation | None:
-        """Return the loki relation if present."""
-        return self.model.get_relation(LOKI_K8S_RELATION)
-
-    @property
-    def grafana_relation(self) -> Relation | None:
-        """Return the grafana relation if present."""
-        return self.model.get_relation(GRAFANA_K8S_RELATION)
+        ops' `get_relation` never returns None for an unknown/`-1`/departing id — it
```
skourta: Can you please explain?

opensearch_single_kernel/core/state.py
```diff
@@ -281,21 +296,40 @@ def sorted_upgrades_units(self) -> list[UpgradeServerState]:
...
+            )
+            return None
+        except RelationNotFoundError:
+            logger.warning(
+                "Remote model unavailable for relation %s, relation is departing.",
+                relation.id,
+            )
+            return None
 
     def peer_cluster_by_relation_id(
```
skourta: How is this different from the one before it?

skourta: Can you also align the arguments with get_peer_cluster_app_model? We should expect the same order.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -37,13 +55,8 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
         self.charm = charm
 
-        self.opensearch_provides = OpenSearchProvides(
-            self.charm,
-            relation_name=CLIENT_RELATION,
-        )
-
         self.framework.observe(
-            self.opensearch_provides.on.index_requested, self._on_index_requested
+            self.charm.state.opensearch_provides.on.resource_requested, self._on_resource_requested
```
skourta: This is very different from the patterns we are using in our charms. Usually we define the providers in the events

opensearch_single_kernel/events/external_clients.py
```diff
@@ -76,132 +86,255 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
             logger.warning(
                 "Modifying relations during an upgrade is not supported."
                 "The charm may be in a broken, unrecoverable state"
             )
         self.charm.external_clients_manager.remove_lingering_relation_users_and_roles(
-            external_client
+            event.relation
         )
 
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
```
skourta: How was this implemented before? This change does not look like related to the interface change. Same for the update status event handler

opensearch_single_kernel/events/external_clients.py
```diff
@@ -76,132 +86,255 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
-                    "index": event.index,
+                    "index": index_name,
                     "id": event.relation.id,
                 },
             ),
             "unit",
             component_name=self.charm.external_clients_manager.name,
         )
 
+        if not self._create_index(event, index_name):
```
skourta: We should use the same pattern all over the charm. Either we return booleans or try and catch.

opensearch_single_kernel/events/external_clients.py
```diff
@@ -76,132 +86,255 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
+            return
+
+        credentials = self._create_user(event, index_name)
+        if not credentials:
+            return
+        username, pwd = credentials
```
skourta: I think the walrus would make it clearer here
```suggestion
        if not (credentials := self._create_user(event, index_name)):
            return
        username, pwd = credentials
```

opensearch_single_kernel/events/external_clients.py
```diff
@@ -76,132 +86,255 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
+
+        updated = False
+        for response in responses:
+            if response.endpoints != new_endpoints:
+                response.endpoints = new_endpoints
+                response.uris = new_endpoints
+                updated = True
+
+        if updated:
+            version = relation.data[relation.app].get("version", "v0") if relation.app else "v0"
```
skourta: Why are we checking on versions. V1 does this automatically. Did we change the model attributes?

opensearch_single_kernel/events/jwt.py
```diff
@@ -51,63 +59,47 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
         ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
             return
 
-        if not self.charm.state.application.is_security_index_initialised:
+        if not self.charm.state.application.security_index_initialised:
             logger.debug("Deferring jwt event as security index not initialised yet")
             event.defer()
             return
 
-        try:
```
skourta: Why remove this?

opensearch_single_kernel/events/jwt.py
```diff
@@ -51,63 +59,47 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
     def _on_jwt_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle relation creation."""
         if (
-            deployment_desc := self.charm.state.application.deployment_desc
+            deployment_desc := self.charm.state.application.deployment_description
         ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
             # Only the main orchestrator applies JWT config.
             logger.warning("JWT relation created on non-main orchestrator.")
 
-    def _on_jwt_relation_changed(self, event: RelationChangedEvent) -> None:
```
skourta: Can you please explain?

opensearch_single_kernel/events/keystore.py
skourta: For other parts we changed it the other way around. We should stick to one pattern throughout the charm

opensearch_single_kernel/events/plugin.py
```diff
@@ -47,17 +46,12 @@ def _on_peer_relation_changed(self, event):  # noqa: C901
...
             # start locally tracking secret and write transferred keys to keystore
-            content = self.charm.state.secrets.get_tracked_secret(
-                plugin.secret_id, Scope.APP, label
-            ).get_content()
-            if not (plugin_config := decode_plugin_secret_content(content, label)):
-                continue
-
-            keys_to_add = plugin_config.get("keys")
+            content = self.charm.plugin_manager.get_plugin_secret(label)
+            keys_to_add = content.get("keys")
```
skourta: Shouldn't we guard on this?

opensearch_single_kernel/core/peer_cluster.py
```diff
@@ -0,0 +1,343 @@
...
+            m.admin_subject = stripped_or_none(source.admin_subject)
+            m.admin_key = stripped_or_none(source.admin_key)
+            m.admin_key_password = stripped_or_none(source.admin_key_password)
+            m.admin_csr = stripped_or_none(source.admin_csr)
+            m.admin_chain = stripped_or_none(source.admin_chain)
+            m.admin_cert = stripped_or_none(source.admin_cert)
+            m.admin_ca_cert = stripped_or_none(source.admin_ca_cert)
+
+            digest_source = source.model_dump(mode="json", context={"skip_secrets": True})
+            m.rel_data_hash = sha1(json.dumps(digest_source, sort_keys=True).encode()).hexdigest()
```
skourta: If we skip the secrets when creating the hash won't the other sub clusters miss the new values?

opensearch_single_kernel/managers/snapshots.py
```diff
@@ -154,7 +157,25 @@ def set_credentials_saved(self, credentials: dict[str, str] | None) -> None:
             del peer_cluster_server.snapshots_credentials_saved
             return
 
-        peer_cluster_server.snapshots_credentials_saved = hash_credentials(credentials)
+        peer_cluster_server.snapshots_credentials_saved = hash_credentials(
+            credentials.model_dump(exclude_none=True)
```
skourta: All credential secret attributes are skipped. The hashes will never match the main orchestrator

opensearch_single_kernel/managers/snapshots.py
```diff
@@ -134,7 +135,9 @@ def read_snapshots_data_from_peer_cluster(
 
         return info_to_save, object_storage_types_to_clean
 
-    def set_credentials_saved(self, credentials: dict[str, str] | None) -> None:
+    def set_credentials_saved(
+        self, credentials: S3RelData | AzureRelData | GcsRelData | None
+    ) -> None:
```
skourta: Shouldn't the `is_provider` be false on line 149?

opensearch_single_kernel/events/peer_cluster.py
```diff
@@ -406,29 +403,29 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+        tls_errors = self.charm.tls_manager.peer_cluster_error_from_tls(remote_peer_cluster)
         self.reconcile_peer_cluster_errors(label="error_from_tls", error=tls_errors)
         if tls_errors:
             logger.debug("TLS/Security misconfigurations detected. Deferring event.")
             event.defer()
             return
 
         # aggregate all CMs (main + failover if any)
-        data.cm_nodes = self.charm.peer_cluster_manager.cm_nodes(orchestrators)
+        remote_peer_cluster.cm_nodes = self.charm.peer_cluster_manager.cm_nodes(orchestrators)
```
skourta: `cm_nodes` is not a field on PeerClusterAppModel

opensearch_single_kernel/events/peer_cluster.py
```diff
@@ -406,29 +403,29 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
-            self.charm.state.application.is_security_index_initialised = True
+        if remote_peer_cluster.security_index_initialised:
+            self.charm.state.application.security_index_initialised = True
 
         # let the charm know this is an already bootstrapped cluster
         self.charm.state.application.bootstrapped = True
         # store the security related settings in secrets, peer_data, disk
-        if data.credentials.admin_tls:
+        if remote_peer_cluster.admin_hashed_password:
             logger.debug("Admin TLS credentials received from peer cluster relation data.")
```
skourta: We update the CA here then do the `peer_cluster_error_from_tls` a few lines below which checks `if self.state.application.admin_ca_cert != peer_cluster_rel_data.admin_ca_cert:` That condition will always be false no?

opensearch_single_kernel/events/jwt.py
```diff
@@ -51,63 +59,47 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
...
         if (
-            deployment_desc := self.charm.state.application.deployment_desc
+            deployment_desc := self.charm.state.application.deployment_description
         ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
             return
 
-        del self.charm.state.server.jwt_auth_configuration
         self.charm.config_manager.update_security_config()
 
         self.apply_security_config_if_needed(event)
```
skourta: We are missing the `authentication_updated` emitted from secretchange of the dplib

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -162,98 +163,111 @@ def _put_relation_user(
...
+        try:
+            tls_ca = self.state.application.admin_chain
+        except KeyError as e:
+            logger.error("Failed to update relation TLS info: missing key %s", str(e))
+            return None
+
+        return {
+            "version": version,
+            "tls_ca": tls_ca,
+            "endpoints": self.get_relation_endpoints(nodes),
```
skourta: what if `get_relation_endpoitns` returns None (line 207)

opensearch_single_kernel/managers/snapshots.py
```diff
@@ -562,113 +574,52 @@ def is_operation_in_progress(self) -> bool:
...
-                (GCS_RELATION, GCS_CREDENTIALS),
-            ]
-            if self.state.secrets.has(Scope.APP, label)
-        ]
-        return [
-            relation_name
-            for relation_name in backup_relations
-            if not self.state.relation_exists(relation_name)
-        ]
+        """Get backup relations that are integrated but missing valid credentials."""
```
skourta: The [status](https://github.com/canonical/opensearch-single-kernel-library/blob/4d4d941100ddb22989f0efe66a356cb965fe11cb/opensearch_single_kernel/managers/snapshots.py#L693) says credentials found but relation missing we are doing the exact opposite here.

skourta (changes requested):
Thank you @Pepsiqqq. This is the first part of the review.
I have some questions notes and some changes that need to be made.
I will do manual testing then review again.

opensearch_single_kernel/events/jwt.py
Pepsiqqq: explain what? just moved it lower to line 78

opensearch_single_kernel/events/jwt.py
Pepsiqqq: it's moved into jwt state [here](https://github.com/canonical/opensearch-single-kernel-library/pull/58/changes/BASE..4d4d941100ddb22989f0efe66a356cb965fe11cb#diff-e08e0519221d8c7ff85c0dec4c19b2162ddd33ebfcddd6a9b89b9ca3a9aefda0R502), we do this because this is needed for config also

opensearch_single_kernel/common/constants.py
Pepsiqqq: Yes, both k8s and vm use 3.8.0

opensearch_single_kernel/core/lock.py
Pepsiqqq: you mean delete lock relation and move fields them to peer?

opensearch_single_kernel/core/state.py
Pepsiqqq: we can but this will create massive circular dependency, we use opensearch provider in events and external/tls managers. So I moved it into state

opensearch_single_kernel/common/client.py
Pepsiqqq: I had some flakiness in provider test, and retry fixed it. Do you want me to remove retry?

opensearch_single_kernel/core/peer_app.py
Pepsiqqq: some of the secret values need to strip white space before comparing with None, logs are readable again. I didn't notice any performance changes

PR #62, opened 2026-04-24
patch(DPE-9760): k8s minor version upgrades (v2)

akram09:
This Pull Request implements the Refresh V2 upgrades for Kubernetes charm. It introduce a specific flow for handling events on kubernetes charm. 
The flow start by charm receiving the `stop` event, meaning the statefulset has already been updated with new revision and pods will start to update one by one (controlled by partition, where pods number bigger than or equal the partition number can update). 
image
Then the charm receives the `upgrade_charm` event where the charm code and workload are already upgraded.
image
Finally, if the upgrade is healthy on bigger unit, the user can resume-upgrade.
image


opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,144 @@
...
+    def client(self) -> Client:
+        """The Lightkube client."""
+        return Client(  # pyright: ignore[reportArgumentType]
+            field_manager=self.pod_name,
+            namespace=self.namespace,
+        )
+
+    # BEGIN: getters
+    def get_ttl_hash(self, seconds=120) -> int:
+        """Gets a unique time hash for the cache, expiring after 2 minutes.
```
carlcsaposs-canonical: would recommend not caching k8s api for refresh v2 usage

refresh v2 assumes response from k8s api is always up-to-date

for example, when setting partition during stop event, v2 implementation only sets partition if stopping unit ordinal is higher than current partition. if unit thinks current partition is higher when it's not (stale cache), then it won't set partition and multiple units can be taken offline at once

carlcsaposs-canonical: also caching will cause delays in detecting if a refresh is in progress, and could cause leader to set partition to 0 since it thinks a refresh is not in progress when it is (taking all units offline at once)

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1084,52 +1098,73 @@ def _post_start_init(self, event: StartOpenSearch) -> None:  # noqa: C901
...
-                # https://github.com/canonical/opensearch-operator/issues/268
-                logger.warning(
-                    "Cluster is yellow. Upgrade may cause data loss if cluster is yellow for reason "
-                    "other than primary shards on upgraded unit & not enough upgraded units available "
-                    "for replica shards"
+            if self.charm.state.substrate == Substrates.K8S:
+                self.charm.state.add_status_if_not_present(
+                    UpgradesStatuses.UPGRADES_WAITING_FOR_RESUME.value,
+                    component="upragdes_manager",
+                    scope="app",
```
carlcsaposs-canonical: status appears to be incorrect if this unit is not the first unit to refresh (e.g. when 2nd or 3rd unit starts after refresh)

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1084,52 +1098,73 @@ def _post_start_init(self, event: StartOpenSearch) -> None:  # noqa: C901
...
+                    # TODO future improvement:
+                    # https://github.com/canonical/opensearch-operator/issues/268
+                    logger.warning(
+                        "Cluster is yellow. Upgrade may cause data loss if cluster is yellow for reason "
+                        "other than primary shards on upgraded unit & not enough upgraded units available "
+                        "for replica shards"
+                    )
+                # Once we are sure that the cluster is healthy, we can mark the unit as healthy
+                if self.charm.state.substrate == Substrates.K8S:
+                    self.charm.state.server_upgrade.unit_state = UnitUpgradesState.HEALTHY
```
carlcsaposs-canonical: can't remember—with v2 you might want to set healthy irrespective of event if applicable (i.e. if unit healthy), like on vm—to improve robustness

carlcsaposs-canonical (commented):
abandoned review part way through—moving functions around makes the diff much harder to review & comment on

(was planning to do a very cursory review [i.e. a very quick glance, not in depth] of all the files, ended up doing a very cursory review of a couple files)

akram09:
Thanks @carlcsaposs-canonical, I addressed your initial comments. Apologies for the state of the PR, we have another PR #64 that this one depends on. Will clean it up and ping you again :ok_hand:

opensearch_single_kernel/events/opensearch.py
```diff
@@ -474,6 +480,14 @@ def _on_install(self, event: InstallEvent) -> None:
         )
         self.charm.workload.install()
 
+    def _on_stop(self, event: StopEvent) -> None:
+        """Event handler for stop event."""
+        self.charm.pebble_observer.stop()
+        if self.charm.substrate == Substrates.K8S and (
+            type(self.charm.upgrades_manager) is UpgradesManagerK8s
```
Gu1nness: nit: Use isinstance instead of `type(x) is X`

opensearch_single_kernel/events/upgrades.py
```diff
@@ -66,9 +73,13 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
             self._on_refresh_force_start_action,
         )
 
+    # -- Event handlers --
+
     def _on_upgrade_peer_relation_created(self, _) -> None:
         """Handle relation created events."""
-        self.charm.upgrades_manager.save_snap_revision_after_first_install()
+        assert self.charm.state.upgrade_relation
```
Gu1nness: This means that you can crash the charm if the peer relation does not exist ?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -122,111 +102,107 @@ def _reconcile_upgrade(self, _=None):  # noqa: C901
...
-        # CHECK FOR ROLLBACK
-        if (
-            self.charm.upgrades_manager.is_rollback
-            and not self.charm.upgrades_manager.can_rollback
-        ):
-            logger.error(
-                "Rollback unsupported. Refresh to a newer revision or consult the recovery documentation"
+    def _upgrade_opensearch(self, event: UpgradeOpenSearch) -> None:  # noqa: C901
+        """Handle upgrade OpenSearch event."""
+        if not (type(self.charm.upgrades_manager) is UpgradesManagerVM):
```
Gu1nness: Same nit here: use isinstance

opensearch_single_kernel/events/upgrades.py
```diff
@@ -279,100 +256,249 @@ def _on_force_upgrade_action(self, event: ops.ActionEvent) -> None:
...
+                f"Rollback status: is_rollback={self.charm.upgrades_manager.is_rollback}, can_rollback={self.charm.upgrades_manager.can_rollback}"
+            )
+            if self.charm.upgrades_manager.is_rollback:
+                logger.warning("Rollback detected")
+                logger.warning(
+                    "Rollback incompatible. Run 'juju run <unit> force-refresh-start' with `check-compatibility` set to false to override node version and attempt startup procedure"
+                )
+                self._set_upgrade_status()
+                return
+            # This is only K8s
```
Gu1nness: If it's the case, then maybe enforce it so that you don't fall in a trap if ever you were to reach here with a VM.

Gu1nness (approved):
I only have tiny nitpicks, nothing blocking.
Congrats !

tests/integration/ha/test_ha_networking.py
Mehdi-Bendriss, line 1: why remove logs from this test suite?

opensearch_single_kernel/charms/base.py
```diff
@@ -162,7 +167,6 @@ def __init__(self, *args):
             self.internal_users_manager,
             self.external_clients_manager,
             self.notifications_manager,
-            self.upgrades_manager,
         )
```
carlcsaposs-canonical, line 170: does this preserve higher/lower priority status ordering?

(IIRC v2 is similar to v3 in this regard; https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/)

opensearch_single_kernel/common/constants.py
```diff
@@ -358,4 +357,5 @@ class ExtraUserRolePermissions(Enum):
 # Upgrades
 UPGRADES_COMPATIBILITY_MATRIX = {
     "2.19.4": {"2.18.0", "2.19.0", "2.19.1", "2.19.2", "2.19.3"},
+    "2.19.5": {"2.19.1", "2.19.4"},
```
carlcsaposs-canonical, line 362: question: why are the ranges so different for 2.19.4 and 2.19.5? (e.g. why is 2.19.1 but not 2.19.2 included for 2.19.5)

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+
+from opensearch_single_kernel.common.exceptions import (
+    OpenSearchK8sDeployedWithoutTrustError,
+)
+
+# default logging from lightkube httpx requests is very noisy
+logging.getLogger("lightkube").disabled = True
+logging.getLogger("lightkube.core.client").disabled = True
+logging.getLogger("httpx").disabled = True
+logging.getLogger("httpcore").disabled = True
```
carlcsaposs-canonical: are you intending to suppress error and warning logs?

consider raising level instead of disabling outright, https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/httpx-logger-level/

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+            if e.status.code == 403:
+                self.on_deployed_without_trust()
+                return
+            raise
+
+    # END: helpers
+
+    # BEGIN: Private methods
+    @cache
+    def _get_partition(self, *_) -> int:
```
carlcsaposs-canonical: https://github.com/canonical/opensearch-single-kernel-library/pull/62#discussion_r3459602700

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+
+        Allows to get cache hit on calls on the same method from different instances of K8sClient
+        as `self` is passed to methods.
+        """
+        return isinstance(other, K8sClient) and self.__dict__ == other.__dict__
+
+    def __hash__(self) -> int:
+        """__hash__ dunder.
+
+        For dict like caching.
```
carlcsaposs-canonical: re cache: https://github.com/canonical/opensearch-single-kernel-library/pull/62#discussion_r3459602700

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+                self.on_deployed_without_trust()
+                return
+            raise
+
+    # END: helpers
+
+    # BEGIN: Private methods
+    @cache
+    def _get_partition(self, *_) -> int:
+        partition = self.client.get(res=StatefulSet, name=self.app_name)
```
carlcsaposs-canonical: I think `--trust` may be needed to get partition as well, but not sure

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+        if not stateful_set.status or not stateful_set.status.updateRevision:
+            raise Exception("Incomplete stateful set")
+        return stateful_set.status.updateRevision
+
+    def _list_revisions(self, *_) -> dict[str, str]:
+        pods = self.client.list(res=Pod, labels={"app.kubernetes.io/name": self.app_name})
+
+        def get_unit_name(pod_name: str) -> str:
+            *app_name, unit_number = pod_name.split("-")
+            return f'{"-".join(app_name)}/{unit_number}'
```
carlcsaposs-canonical:
```suggestion
            app_name, unit_number = pod_name.rsplit("-", maxsplit=1)
            return f'{app_name)/{unit_number}'
```
nit

opensearch_single_kernel/common/statuses.py
```diff
@@ -329,6 +329,16 @@ class UpgradesStatuses(Enum):
...
         approved_critical_component=True,
     )
+    K8S_UPGRADES_ACTIVE = StatusObject(
+        status="active",
+        message="OpenSearch {workload_version} running; Charmed operator {charm_version}",
+        approved_critical_component=True,
+    )
+    K8S_UPGRADES_ACTIVE_OUTDATED = StatusObject(
+        status="active",
+        message="OpenSearch {workload_version} running (outdated); Charmed operator {charm_version}",
```
carlcsaposs-canonical: recommend (restart pending) instead of (outdated) for k8s only because of https://bugs.launchpad.net/juju/+bug/2036246

example: https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#kubernetes-rollback

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+        """Get unit upgrade status."""
+        if not self.state.upgrade_relation:
+            return (None, None)
+        if self.is_rollback:
+            if not self.can_rollback:
+                return UpgradesStatuses.UPGRADES_ROLLBACK_UNSUPPORTED.value, None
+            else:
+                return UpgradesStatuses.UPGRADES_ROLLBACK_INCOMPATIBLE.value, {
+                    "param": "check-workload-container"
+                }
```
carlcsaposs-canonical: if workload digest != original workload digest, suggest not considering that a rollback (i.e. in a rollback case, check-workload-container=false should never be applicable)

opensearch_single_kernel/common/statuses.py
```diff
@@ -361,6 +376,6 @@ class UpgradesStatuses(Enum):
     )
     UPGRADES_ROLLBACK_INCOMPATIBLE = StatusObject(
         status="blocked",
-        message="Rollback incompatible. Run 'juju run <unit> force-refresh-start' with `check-compatibility` set to false to override node version and attempt startup procedure",
+        message="Rollback incompatible. Run 'juju run <unit> force-refresh-start' with `{param}` set to false to override node version and attempt startup procedure",
```
carlcsaposs-canonical: suggest reverting, https://github.com/canonical/opensearch-single-kernel-library/pull/62/changes#r3497586681

opensearch_single_kernel/core/state.py
```diff
@@ -254,14 +254,24 @@ def sorted_server_upgrades(self) -> list[UpgradeServerState]:
...
 
+    @property
+    def pod_name(self) -> str:
+        """K8S only: The pod name."""
+        return self.model.unit.name.replace("/", "-")
+
+    @property
+    def namespace(self) -> str:
+        """K8S only: The namespace."""
+        return self.model.unit._backend.model_name
```
carlcsaposs-canonical:
```suggestion
        return self.model.name
```
avoid using ops' private api

opensearch_single_kernel/events/opensearch.py
```diff
@@ -467,6 +473,14 @@ def _on_install(self, event: InstallEvent) -> None:
...
+        ):
+            self.charm.upgrades_manager.prepare_for_shutdown()
```
carlcsaposs-canonical, line 483: recommend setting partition as very first thing in stop event (i.e. before starting graceful shutdown of workload)

if we fail to set partition, multiple units can go offline—during stop event, setting partition is highest priority

(reason for setting first is to ensure it has time to run before pod is terminated [usually 30 seconds, but can be less] & in case other things during stop event fail)

opensearch_single_kernel/events/opensearch.py
```diff
@@ -782,6 +796,13 @@ def _on_start(self, event: StartEvent) -> None:  # noqa: C901
 
     def _on_start_opensearch(self, event: StartOpenSearch) -> None:  # noqa: C901
         """Start OpenSearch, with a generated or passed conf, if all resources configured."""
+        if (
+            not event.after_upgrade
+            and self.charm.substrate == Substrates.K8S
+            and self.charm.upgrades_manager.is_rollback
```
carlcsaposs-canonical, line 807: question: when would `event.after_upgrade` be False and `self.charm.upgrades_manager.is_rollback` be True?

opensearch_single_kernel/managers/peer_cluster.py
```diff
@@ -446,11 +449,11 @@ def get_statuses(  # noqa: C901
                     )
             for peer_cluster in self.state.peer_clusters(remote=True, is_provider=False):
                 # check if there is an error
-                if (error_data := peer_cluster.error_data) and (status := error_data.get_status()):
-                    status_list.append(status)
+                if error_data := peer_cluster.error_data:
+                    status_list.append(error_data.get_status())
```
carlcsaposs-canonical: fyi looks like return type for `get_status()` can be None

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+        return self.k8s_client.get_partition()
+
+    @partition.setter
+    def partition(self, value: int) -> None:
+        """Sets the partition number."""
+        self.k8s_client.set_partition(value)
+
+    def save_revision_after_first_install(self) -> None:
+        """Save revision on first install"""
+        self.state.server_upgrade.workload_version = self.current_versions.workload
```
carlcsaposs-canonical: recommend saving charm version, workload version, and workload digest (to reliably detect rollback)

carlcsaposs-canonical: or perhaps that is set elsewhere and this was mistakenly copied from vm? (since there's no workload installation on k8s)

opensearch_single_kernel/managers/upgrades_k8s.py
carlcsaposs-canonical: i.e. suggest using "rollback" as defined by https://canonical-charm-refresh.readthedocs-hosted.com/latest/glossary/#app (I believe the v2 design uses the same definition for rollback)

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+                },
+            )
+        else:
+            return (
+                UpgradesStatuses.K8S_UPGRADES_ACTIVE_OUTDATED.value,
+                {
+                    "workload_version": self.current_versions.workload,
+                    "charm_version": self.current_versions.charm,
+                },
+            )
```
carlcsaposs-canonical, line 116: might be worth caching this api call for duration of juju event

(1. to reduce calls to k8s api. 2. to ensure consistency in charm logic [i.e. so that the values aren't changing when called in different places])

the partition is what's important not to cache; statefulset controller revision hash is okay to cache

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+            return (
+                UpgradesStatuses.K8S_UPGRADES_ACTIVE.value,
+                {
+                    "workload_version": self.current_versions.workload,
+                    "charm_version": self.current_versions.charm,
+                },
+            )
+        else:
+            return (
+                UpgradesStatuses.K8S_UPGRADES_ACTIVE_OUTDATED.value,
```
carlcsaposs-canonical, line 111: https://github.com/canonical/opensearch-single-kernel-library/pull/62/changes#r3497560174

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+        Automatically upgrades next unit if all upgraded units are healthy—except if only one unit
+        has upgraded (need manual user confirmation [via Juju action] to upgrade next unit)
+
+        Handle Juju action to:
+        - confirm first upgraded unit is healthy and resume upgrade
+        - force upgrade of next unit if 1 or more upgraded units are unhealthy
+        """
+        message: str | None = None
+        force = action_event is not None and force
+        from_event = action_event is not None
```
carlcsaposs-canonical:
```suggestion
```
suggest using `action_event` directly instead; this is confuisng

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+
+        Automatically upgrades next unit if all upgraded units are healthy—except if only one unit
+        has upgraded (need manual user confirmation [via Juju action] to upgrade next unit)
+
+        Handle Juju action to:
+        - confirm first upgraded unit is healthy and resume upgrade
+        - force upgrade of next unit if 1 or more upgraded units are unhealthy
+        """
+        message: str | None = None
+        force = action_event is not None and force
```
carlcsaposs-canonical: why is force not determined from the action parameter?

force in this context refers only to check-health-of-refreshed-units=false (https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/actions/#check-health-of-refreshed-units-false). it does not refer to force-refresh-start

opensearch_single_kernel/events/opensearch.py
```diff
@@ -1054,52 +1075,72 @@ def _post_start_init(self, event: StartOpenSearch) -> None:  # noqa: C901
...
-                # https://github.com/canonical/opensearch-operator/issues/268
-                logger.warning(
-                    "Cluster is yellow. Upgrade may cause data loss if cluster is yellow for reason "
-                    "other than primary shards on upgraded unit & not enough upgraded units available "
-                    "for replica shards"
+            # Once we are sure that the cluster is healthy, we can mark the unit as healthy
+            if self.charm.state.substrate == Substrates.K8S:
+                if self.charm.upgrades_manager.is_rollback:
+                    # This means we run force-refresh-start to override version
+                    self.charm.upgrades_manager.reconcile_partition(action_event=event, force=True)
```
carlcsaposs-canonical: why is force=True set here?

https://github.com/canonical/opensearch-single-kernel-library/pull/62/changes#r3498024614

opensearch_single_kernel/events/upgrades.py
```diff
@@ -279,30 +320,36 @@ def _on_force_upgrade_action(self, event: ops.ActionEvent) -> None:
...
-        # DA091 approved
-        # (https://docs.google.com/document/d/1rwnS-deJU9Mzc8BFkl3UGgjZiBa6e3bxoT-6BQo9e3E/edit)
-        self.charm.upgrade_opensearch_event.emit(ignore_lock=False)
+        if self.charm.substrate == Substrates.VM:
+            # TODO: replace `ignore_lock=False` with `event.params["ignore-lock"]` if specification
+            # DA091 approved
+            # (https://docs.google.com/document/d/1rwnS-deJU9Mzc8BFkl3UGgjZiBa6e3bxoT-6BQo9e3E/edit)
+            self.charm.upgrade_opensearch_event.emit(ignore_lock=False)
+        else:
+            self.charm.upgrades_manager.reconcile_partition(action_event=event, force=True)
```
carlcsaposs-canonical: on k8s, the partition should be reconciled by the leader

this is why v2 does not use the force-upgrade action on k8s, and instead adds a parameter to resume-upgrade. force-upgrade does not map to k8s

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+            logger.debug(f"Partition set to {current_unit_number} during stop event")
+
+        if not self.state.upgrade_relation:
+            logger.debug("Upgrade Peer relation missing during stop event")
+            return
+
+        try:
+            self.opensearch_client.disable_shard_allocation(alt_hosts=self.alt_hosts)
+        except OpenSearchHttpError:
+            logger.exception("Failed to disable shard allocation before upgrade")
```
carlcsaposs-canonical, line 217: I believe this runs during any pod eviction, not just before upgrade

opensearch_single_kernel/events/upgrades.py
```diff
@@ -79,19 +88,20 @@ def _on_upgrade_peer_relation_created(self, _) -> None:
...
-            self.charm.upgrades_manager.current_versions,
-        )
-        self.charm.state.application_upgrade.versions = (
-            self.charm.upgrades_manager.current_versions
-        )
-        logger.debug(
-            "Set %r in upgrade peer relation app databag",
-            self.charm.upgrades_manager.current_versions,
-        )
+        if self.charm.unit.is_leader():
```
carlcsaposs-canonical: is this not already checked above?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -261,14 +299,17 @@ def _on_resume_upgrade_action(self, event: ops.ActionEvent) -> None:
...
-
-        self.charm.upgrades_manager.reconcile_partition(action_event=event)
+        try:
+            self.charm.upgrades_manager.reconcile_partition(action_event=event)
+        except OpenSearchReconcilePartitionError as e:
+            logger.debug(f"Resume upgrade event failed: {e}")
+            event.fail(str(e))
+            return
         # If next to upgrade, upgrade leader unit
         self._reconcile_upgrade()
```
carlcsaposs-canonical: is it intentional that reconcile_partition is being called multiple times here?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -355,24 +402,46 @@ def _on_refresh_force_start_action(self, event: ops.ActionEvent) -> None:
...
+            self.charm.substrate == Substrates.VM
+            and self.charm.upgrades_manager.unit_state is not UnitUpgradesState.OUTDATED
+        ):
             message = "Unit already upgraded"
             logger.debug(f"Force upgrade event failed: {message}")
             event.fail(message)
             return
-
-        if event.params.get("check-compatibility", True):
+        if self.charm.substrate == Substrates.VM and event.params.get("check-compatibility", True):
```
carlcsaposs-canonical: why are these checks being skipped on k8s?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -355,24 +402,46 @@ def _on_refresh_force_start_action(self, event: ops.ActionEvent) -> None:
...
-        )
-        logger.debug("Overrode OpenSearch version")
+        if self.charm.substrate == Substrates.K8S and event.params.get(
+            "check-workload-container", True
+        ):
+            # Cannot force refresh start if workload container image check fails
+            message = "Cannot force refresh start if workload container image check fails. Rerun with `check-workload-container` set to false to override this check and attempt startup procedure."
+            logger.debug("Refresh force start event failed: %s", message)
+            event.fail(message)
+            return
```
carlcsaposs-canonical: why does the event always fail if check-workload-container is True?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -432,3 +501,52 @@ def authorized_leader(self) -> bool:
...
+            event.defer()
+            return
+
+        # Mark the new version of the unit since in Kubernetes this unit is upgraded now.
+        self.charm.state.server_upgrade.workload_version = (
+            self.charm.upgrades_manager.current_versions.workload
+        )
+        logger.debug(
+            f"Saved {self.charm.upgrades_manager.current_versions.workload=} in unit databag after upgrade"
+        )
```
carlcsaposs-canonical, line 572: question: what is this for?

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+    def get_revision(self) -> str:
+        """Gets the stateful set revision."""
+        return self._get_revision()
+
+    def list_revisions(self) -> dict[str, str]:
+        """Returns a mapping of {unit name: Kubernetes controller revision hash.
+
+        This is used for kubernetes upgrades to get the version of each container.
+        """
+        return self._list_revisions()
```
Mehdi-Bendriss: these are simply proxy methods, they're not needed. You can simply expose the ones these call.

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,133 @@
...
+            raise Exception("Incomplete stateful set.")
+        return partition.spec.updateStrategy.rollingUpdate.partition
+
+    def _get_revision(self, *_) -> str:
+        stateful_set = self.client.get(res=StatefulSet, name=self.app_name)
+        if not stateful_set.status or not stateful_set.status.updateRevision:
+            raise Exception("Incomplete stateful set")
+        return stateful_set.status.updateRevision
+
+    def _list_revisions(self, *_) -> dict[str, str]:
```
Mehdi-Bendriss: A few points here:
1. can you add docstrings to all these methods
2. don't raise `Exception`, instead raise a specific exception
3. are those exceptions caught somewhere?

carlcsaposs-canonical (commented):
(only took a quick glance; did not do a thorough review)

from looking at the implementation, it seems like there are some foundational misunderstandings of how upgrades v2 works/is designed

(I know I've mentioned this before and it sounds like a decision has already been made, but repeating here for history:) given that opensearch k8s does not yet have a stable release (and the backwards compatibility requirements that come with it), I think it might be significantly easier to implement refresh v3 for k8s (while keeping vm on v2 in the same codebase) than to implement k8s on v2. I recognize that having v2 and v3 in the same codebase would be quite a hassle, difficult to comprehend, and difficult to implement correctly. With that said, I still think it would be more difficult to implement a robust implementation with v2 than with v3. (given that v2 does not have a clear api while v3 has a clear api & docs. also FYI you'll be the only charm on v2 for k8s [the only other historical one was mysql-router-k8s, but that's on v3 now])

fyi, skipped review on large parts of events/opensearch.py, events/upgrades.py, and managers/upgrade_k8s.py after realizing there were foundational misunderstandings. also did not review anything in tests/

opensearch_single_kernel/core/state.py
```diff
@@ -587,10 +597,10 @@ def fqdn(self) -> str:
     @property
     def network_hosts(self) -> list[str]:
         """All HTTP/Transport hosts for the current node."""
-        hosts = ["_site_", "_local_"]
+        hosts = ["_site_"]
         if self.substrate == Substrates.K8S:
             # K8s we allow binding on all interfaces
-            return hosts + [self.fqdn]  # only the DNS name that's in the cert SANs
+            return hosts + [self.fqdn] + ["_local_"]  # only the DNS name that's in the cert SANs
```
Mehdi-Bendriss: I thought we had removed `_local_` from the network.hosts and from the sans_dns in the TLS module ?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -782,6 +796,13 @@ def _on_start(self, event: StartEvent) -> None:  # noqa: C901
 
     def _on_start_opensearch(self, event: StartOpenSearch) -> None:  # noqa: C901
         """Start OpenSearch, with a generated or passed conf, if all resources configured."""
+        if (
+            not event.after_upgrade
+            and self.charm.substrate == Substrates.K8S
+            and self.charm.upgrades_manager.is_rollback
+        ):
+            event.defer()
```
Mehdi-Bendriss, line 809: can we add a log before the deferral?

opensearch_single_kernel/events/upgrades.py
Mehdi-Bendriss: I think so with `if not self.authorized_leader:`

opensearch_single_kernel/events/upgrades.py
```diff
@@ -79,19 +88,20 @@ def _on_upgrade_peer_relation_created(self, _) -> None:
...
+            self.charm.state.application_upgrade.versions = (
+                self.charm.upgrades_manager.current_versions
+            )
+            logger.debug(
+                "Set %r in upgrade peer relation app databag",
+                self.charm.upgrades_manager.current_versions,
+            )
 
-    def _reconcile_upgrade(self, _=None):  # noqa: C901
+    def _reconcile_upgrade(self, during_upgrade=False):  # noqa: C901
```
Mehdi-Bendriss: nit, missing type hint on arg and return type

opensearch_single_kernel/events/upgrades.py
```diff
@@ -167,6 +185,35 @@ def _reconcile_upgrade(self, _=None):  # noqa: C901
...
+                f"Rollback status: is_rollback={self.charm.upgrades_manager.is_rollback}, can_rollback={self.charm.upgrades_manager.can_rollback}"
+            )
+            if self.charm.upgrades_manager.is_rollback:
+                logger.warning("Rollback detected")
+                logger.warning(
+                    "Rollback incompatible. Run 'juju run <unit> force-refresh-start' with `check-compatibility` set to false to override node version and attempt startup procedure"
+                )
+                self._set_upgrade_status()
+                return
+            # This is only K8s
```
Mehdi-Bendriss: nit, could you add a line break before the comment?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -167,6 +185,35 @@ def _reconcile_upgrade(self, _=None):  # noqa: C901
...
+                logger.info(
+                    "Refresh incompatible. If you accept potential *data loss* and *downtime*, you can continue with force-upgrade"
+                )
+                self.charm.state.add_status_if_not_present(
+                    UpgradesStatuses.UPGRADES_INCOMPATIBLE.value,
+                    "unit",
+                    self.charm.upgrades_manager.name,
+                )
+                return
+        if self.charm.state.substrate == Substrates.K8S:
```
Mehdi-Bendriss, line 175: nit, could you add a line break before the comment?

opensearch_single_kernel/events/upgrades.py
```diff
@@ -235,13 +275,11 @@ def _on_pre_upgrade_check_action(self, event: ops.ActionEvent) -> None:
             logger.debug(f"Pre-upgrade check event failed: {message}")
             event.fail(message)
             return
-
```
Mehdi-Bendriss, line 238: could you restore the line breaks in all this file (multiple occurrences), as the code becomes too crammed otherwise and hard to read

opensearch_single_kernel/events/upgrades.py
Mehdi-Bendriss: nit, line break please

opensearch_single_kernel/common/constants.py
```diff
@@ -326,7 +326,6 @@ class ExtraUserRolePermissions(Enum):
 DATA_ROLE_REMOVAL_FORBIDDEN = (
     "Removal of data role from current deployment not allowed - the data cannot be reallocated."
 )
-
```
Mehdi-Bendriss, line 329: line break, please 🥲

opensearch_single_kernel/events/custom_events.py
```diff
@@ -23,29 +23,25 @@ def __init__(
         ignore_lock: bool = False,
         after_upgrade: bool = False,
         is_first_data_node: bool = False,
-        override_version: bool = False,
```
Mehdi-Bendriss, line 26: can you explain the removal of this?

opensearch_single_kernel/events/opensearch.py
```diff
@@ -687,7 +702,6 @@ def _on_start(self, event: StartEvent) -> None:  # noqa: C901
             )
             event.defer()
             return
-
```
Mehdi-Bendriss, line 686: nit, line break

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+        https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions
+
+        For Kubernetes, unit numbers are guaranteed to be sequential.
+        """
+        return self.k8s_client.get_partition()
+
+    @partition.setter
+    def partition(self, value: int) -> None:
+        """Sets the partition number."""
+        self.k8s_client.set_partition(value)
```
Mehdi-Bendriss, line 67: are the propagated exceptions handled in the events handler?

opensearch_single_kernel/managers/upgrades_k8s.py
```diff
@@ -0,0 +1,265 @@
...
+    def save_revision_after_first_install(self) -> None:
+        """Save revision on first install"""
+        self.state.server_upgrade.workload_version = self.current_versions.workload
+        logger.debug(
+            f"Saved {self.current_versions.workload=} in unit databag after first install"
+        )
+
+    @property
+    def unit_status(self) -> tuple[StatusObject | None, dict[str, Any] | None]:
+        """Get unit upgrade status."""
```
Mehdi-Bendriss, line 84: can you explain the return type? what the tuple represents

opensearch_single_kernel/managers/upgrades_vm.py
```diff
@@ -42,14 +42,18 @@ def unit_state(self) -> UnitUpgradesState | None:
     @property
     def unit_status(self) -> tuple[StatusObject | None, dict[str, Any] | None]:
         """Get unit upgrade status."""
```
Mehdi-Bendriss, line 44: similar comment on the docstrings

Mehdi-Bendriss (commented):
Thanks Akram! I left small comments on the code itself, since Carl already walked through the upgrades logic / robustness

opensearch_single_kernel/common/constants.py
akram09: As discussed in the call. These are the current versions I tested rollbacks to we will be adding more once manual testing is done

opensearch_single_kernel/events/upgrades.py
akram09: Fixed, now the resume-upgrade on k8s have force param and this action runs only on VM

opensearch_single_kernel/managers/upgrades_k8s.py
akram09: Fixed, thank you.

opensearch_single_kernel/events/upgrades.py
akram09: Fixed, Thank you

opensearch_single_kernel/events/upgrades.py
akram09: Fixed, Thank you

opensearch_single_kernel/events/upgrades.py
akram09: We are now only using the check-compatibility param as discussed in the call.

tests/integration/ha/test_ha_networking.py
akram09: These changes are coming from the LD-K8s PR (will be fixed once I merge again with that PR)

opensearch_single_kernel/common/k8s.py
akram09: Right, it used to be useful for caching but we don't do that anymore

tests/integration/ha/test_ha_networking.py
Mehdi-Bendriss: okay, please only merge once Carl has re-reviewed, so that the diff is constrained to the exact scope of upgrades

opensearch_single_kernel/charms/base.py
akram09: This make sure that upgrades Statuses are the highest priority of the charm statuses

carlcsaposs-canonical: some statuses should have higher priority, while others should have lower priority

for example, if unit is unhealthy after upgrade, the status indicating what is unhealthy should take priority over the lower priority upgrade status

akram09: Since we are using advanced status, we can order them in the `get_statuses` function

opensearch_single_kernel/events/upgrades.py
akram09: The Restarting state is only used in K8s

opensearch_single_kernel/events/custom_events.py
akram09: This fixes one of the regressions we found which is that we mistakenly added override_version to the startOpenSearchEvent and not the UpgradeOpenSearch Event

opensearch_single_kernel/managers/upgrades_k8s.py
akram09: I removed the exceptions and added them in one place where we check if we have trust enabled.

opensearch_single_kernel/managers/upgrades_k8s.py
akram09: The return type here is StatusObject and the params that can be used to format the status.

opensearch_single_kernel/charms/base.py
carlcsaposs-canonical: by higher/lower priority, I mean relative to the other statuses in the charm (not relative within the upgrade statuses)

carlcsaposs-canonical: ^

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,116 @@
...
+from lightkube.resources.authorization_v1 import SelfSubjectAccessReview
+from lightkube.resources.core_v1 import Pod
+
+from opensearch_single_kernel.common.exceptions import (
+    OpenSearchK8sDeployedWithoutTrustError,
+)
+
+# default logging from lightkube httpx requests is very noisy
+logging.getLogger("lightkube").disabled = True
+logging.getLogger("lightkube.core.client").disabled = True
```
carlcsaposs-canonical: https://github.com/canonical/opensearch-single-kernel-library/pull/62#discussion_r3497460210

opensearch_single_kernel/common/k8s.py
```diff
@@ -0,0 +1,116 @@
...
+        ):
+            logger.warning(
+                f"Run `juju trust {self.app_name} --scope=cluster`. Needed for in-place refreshes"
+            )
+            raise OpenSearchK8sDeployedWithoutTrustError(app_name=self.app_name)
+
+    def on_deployed_without_trust(self) -> None:
+        """Blocks the application and returns a specific error message."""
+        logger.error("Kubernetes application needs `juju trust`")
+        raise OpenSearchK8sDeployedWithoutTrustError(app_name=self.app_name)
```
carlcsaposs-canonical: appears not to be called anywhere, is this redundant?

opensearch_single_kernel/common/statuses.py
```diff
@@ -342,17 +352,27 @@ class UpgradesStatuses(Enum):
     )
     UPGRADES_WAITING_FOR_RESUME = StatusObject(
         status="blocked",
-        message="Upgrading. Verify highest unit is healthy & run `resume upgrade action.",
+        message="Upgrading. Verify highest unit is healthy & run `resume-upgrade` action.",
+        approved_critical_component=True,
+    )
+    K8S_UPGRADES_WAITING_FOR_RESUME = StatusObject(
+        status="blocked",
+        message="Upgrading. Verify highest unit is healthy & run `resume-refresh` action.",
```
carlcsaposs-canonical: question: why are the actions using different names on vm/k8s?

opensearch_single_kernel/managers/upgrades_base.py
```diff
@@ -235,37 +292,51 @@ def reconcile_compatibility_matrix(self) -> dict[str, set[str]]:
...
+            f"{self.state.server_upgrade.unit.name=} {self.state.sorted_upgrades_units[0].unit.name=} {self.in_progress=} {self.is_compatible=} {self.state.application_upgrade.versions.workload_parsed=} {self.current_versions.workload_parsed=} {self.get_version_before_override()=}"
+        )
+        return (
+            self.state.server_upgrade.unit.name == self.state.sorted_upgrades_units[0].unit.name
+            and self.in_progress
+            and self.is_compatible
+            and self.state.application_upgrade.versions.workload_parsed
+            == self.current_versions.workload_parsed
+            and self.get_version_before_override() is not None
+        )
```
carlcsaposs-canonical: to avoid bugs, would recommend not checking
- if this is highest unit
- if refresh in progress
- if refresh compatible
in is_rollback—I believe you would want to check those conditions before checking if it is a rollback

https://chat.canonical.com/canonical/pl/77b4nr1zg7dn9m7yay68hnyuxo

also, it appears get_version_before_override currently can return None if there's an error when calling the command—making it possible for this function to return False when a rollback is in progress

opensearch_single_kernel/charms/base.py
akram09: The upgrades statuses are configured to have the highest priority compared to other statuses (Recommended by Mykola)

opensearch_single_kernel/common/k8s.py
akram09: I thought you were recommending this only to httpx

opensearch_single_kernel/common/k8s.py
akram09: Yes this is leftover. Thanks

opensearch_single_kernel/common/statuses.py
akram09: It was requested that we try to align to naming conventions of v3 in K8s

opensearch_single_kernel/managers/upgrades_base.py
akram09: The thing is we don't only check the rollback in one hook, we block the start and startopenSearch hooks if it is a rollback so that the charm is not stuck in trying to start with a rolled back version. 

I would personally, not have all these checks each time want to check if its rollbacks.

akram09: On the second comment, I just improved the robustness of that function. Thank you

opensearch_single_kernel/events/upgrades.py
```diff
@@ -36,6 +44,9 @@ class UpgradesEventsHandler(Object):
 
     lifecycle_state_stored = ops.StoredState()
 
+    # This is to differentiate between substrates
+    UPGRADE_NOTATION = "refresh" if Substrates.K8S else "upgrade"
+
     def __init__(self, charm: "OpenSearchBaseCharm") -> None:
         super().__init__(charm, key="upgrade_events")
         self.charm = charm
```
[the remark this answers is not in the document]
akram09: It is okey since the substrate is a fixed value per charm

PR #82, opened 2026-05-21
patch: Bugfixes

### LINKED ISSUE #65: Fix return type of `reconcile_peer_relation_data`
The current return type of `reconcile_peer_relation_data` is a boolean that states if the event should be defered or not. We don't want this actually since The managers should be more agnostic on the details of events handling

### LINKED ISSUE #66: A demoted main orchestrator still overrides the orchestration data after it has been demoted.
We demote (old) main when we re-relate it to the new orchestrator (named failover) if we do not do it and instead relate it again to data nodes it keep on thinking it is the main orchestrator and will cause an incoherence for the data nodes. We should block this from happening by also sendinf the promotions times in a future PR.

Image

### LINKED ISSUE #75: Status "blocked: cluster_manager and voting_only cannot be both set" cleared without fixed
## Steps to reproduce

Deploy opensearch and set the roles to `cluster_manager, voting_only` A status will be set but after a set of hooks it will be removed while the role is still invalid.

## Actual behavior
The status should be persisted.

### LINKED ISSUE #76: "app-admin.p12" is not stored after remove and integrate with tls
## Steps to reproduce

1. Deploy opensearch with 3 units 
2. Relate to TLS 
3. Remove the TLS relation 
4. Relate again the TLS relation  
5. Repeat the 3-4 another time

## Actual behavior
The units except leader will be stuck with a "Waiting for TLS" status because the "app-admin.p12" is not stored in the non-leader units.

s-mlnk:
Fixes #65: return type was fixed and inverted to logically decouple it from event handling. TLDR: "return whether should be deferred" -> "return whether operation was successful".

Fixes #66: issue arise from two main orchestrators being linked to one subcluster. Now the check on this case is fixed.

Fixes #75: issue arise from check responsible for `cluster_manager and voting_only roles cannot be both set on the same node` status being logically unreachable. Now the checks are reshuffled to ensure correct order.

Fixes #76: issue arise from when subordinate units receive a new certs before the leader does, while the old certificate (e.g. from previous relation) is still in admin secret, which resulted in old cert <> new private key conflict. A new defer check is added for that case.


opensearch_single_kernel/events/tls.py
```diff
@@ -368,7 +368,9 @@ def on_tls_conf_set(
             if not admin_secrets.get("subject"):
                 return
 
-        self.charm.tls_manager.store_admin_tls_secrets_if_applies()
+        if not self.charm.tls_manager.store_admin_tls_secrets_if_applies():
```
akram09, line 371: Why does this fail in the case of CA rotation ? I asking since maybe this is not the right place to defer

opensearch_single_kernel/managers/peer_cluster.py
```diff
@@ -145,7 +145,14 @@ def reconcile_orchestrators_from_provider_data(
 
         local_orchestrators = self.state.application.orchestrators_dict
 
-        if (trigger in {"main", "failover"}) and (relation_units > 0):
+        if (
```
akram09: Can you provide more context on why this change ?

akram09 (commented):
Thank you Serhii, seems good for me. 
Regarding the issue of roles, I think I might misexplained the bug in the issue. Basically, whenever we have `cluster_manager` and `voting_only` roles set on the same node we need to block and say that a node can not have both roles in the same node. Which is not really the case when I tested this PR. There is also the thing that we are only adding the status when we compute in `ClusterManager._existing_cluster_setup`

opensearch_single_kernel/managers/peer_cluster.py
s-mlnk: Previously this function was adding values in dict without any check on overriding. For example, when the second main orchestrator is connected, it just overrides a `main_app` & `main_rel_id` values with the new ones. But it shouldn't be if we actually have an active connection to main orchestrator, and it also makes this check broken:
https://github.com/canonical/opensearch-single-kernel-library/blob/6462da7916ccea2b7ffa32ce0f23f1b6276aefef/opensearch_single_kernel/managers/peer_cluster.py#L234-L238
(this code checks "if newly added relation is not within an active connections list, then it's place already taken and we signalise about conflict")

skourta: Can you please further explain @s-mlnk. We should not be able to connect to more than one main orchestrator.

opensearch_single_kernel/events/tls.py
skourta: Can you please explain here again. I believe this was tackled on a sync call

skourta (commented):
Couple of questions otherwise LGTM. 
Thank you @s-mlnk

opensearch_single_kernel/managers/peer_cluster.py
s-mlnk: We can have a look on `_on_peer_cluster_relation_changed` in `events/peer_cluster.py:261`:
1. line 313: we compute a new `orchestrators` dictionary with main & failover orchestrators info **along with the new relation assigned**.
    It means that if we receive second "main orchestrator" connection, we will mistakenly override our current "main orchestrator" relation here.
2. line 383: we already had a check in `reconcile_peer_cluster_errors` function specifically for the "this relation is not among active connections in `orchestrators` dict, so it should be conflict here" situation!
    But this check will never work because we always attach new relation into the `orchestrators` dict.
3. line 389: if the check above would've worked, the peer relation event would've exited here and new relation would've been ignored with the correct error message.

The same bug happens on OG OpenSearch with the same logic. From the existence of that check, I think it was the intended way of handling this error from the very beginning. And even from the modern perspective, I think this check is still in right place as it's early enough to prevent any changes made from new relation.

opensearch_single_kernel/events/tls.py
s-mlnk: image

Why can't we just erase the old admin certificate during certificate request? Because there's actually a check for TLS renewal using the admin cert. If we remove it, the CA rotation then breaks. My suggestion is to hot-fix this issue until the TLS V4 is established and whole process is refactored so no time is wasted.

s-mlnk:
@akram09 I finally understood what the #75 was and added a fix here.

opensearch_single_kernel/managers/peer_cluster.py
skourta: You have not cleared these @s-mlnk

s-mlnk: I've already moved this check out but forgot to clean it up there. Now it's done

opensearch_single_kernel/core/models.py
```diff
@@ -836,6 +836,12 @@ def promote_failover(self) -> None:
         self.main_rel_id = self.failover_rel_id
         self.delete("failover")
 
+    def check_relation_conflict(self, trigger: str, relation_id: int) -> bool:
```
akram09, line 839: nit: Rename to `is_trigger_conflicting`

opensearch_single_kernel/events/peer_cluster.py
```diff
@@ -701,3 +712,26 @@ def _reconcile_deployment_desc_from_peer_cluster_data(self, data: PeerClusterRel
         self.charm.opensearch_events.apply_status_from_deployment_desc(
             self.charm.state.application.deployment_desc
         )
+
+    def _put_relation_conflict_error(
```
akram09: On Kubernetes PR we are not really using the "errros-from-providers" value we just set the status on get_statuses so I would say we just add the status at that step instead of adding error to the databag

opensearch_single_kernel/events/opensearch.py
```diff
@@ -640,7 +640,6 @@ def _on_start(self, event: StartEvent) -> None:  # noqa: C901
             logger.debug("Blocking directives present. Deferring start event.")
             event.defer()
             return
-        self.charm.cluster_manager.clear_directive(Directive.SHOW_STATUS)
```
akram09, line 677: When I tested this PR, it seems that the new roles values are rejected but we don't show any blocked status. (The config is set to "")

akram09 (approved):
Thank you Serhii, Seems good

opensearch_single_kernel/events/peer_cluster.py
s-mlnk: I made it the same as other errors. I think we should leave it just to keep things uniform.

opensearch_single_kernel/events/opensearch.py
Mehdi-Bendriss: An wrong role change should result in a blocked status. 
Is this no longer showing due the removal of the `SHOW_STATUS` directive?

akram09 (changes requested):
Hello @s-mlnk, I tested again your PR. It seems good, apart from the fact that when we set the new roles OpenSearch should not restart to apply these roles but refrain from updating the config. 

The blocked status is now shown but we need to make sure that config is not updated since OpenSearch will not be allowed to work with these roles.

opensearch_single_kernel/events/opensearch.py
s-mlnk: @akram09 confirmed it now works.

Mehdi-Bendriss (dismissed):
Thanks Serhii! Looks good to me (pending CI pass)

PR #97, opened 2026-06-28
patch: LDAP integration

s-mlnk:
## How was this change tested?
- [x] Manually
- [ ] Unit tests
- [x] Integration tests

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.


opensearch_single_kernel/managers/config.py
```diff
@@ -434,6 +411,72 @@ def _security_authc_oauth_config(self) -> dict[str, Any]:
...
+                        "type": "basic",
+                        "challenge": False,
+                    },
+                    "authentication_backend": {
+                        "type": "ldap",
+                        "config": {
+                            "enable_ssl": True,
+                            "enable_start_tls": False,
+                            "enable_ssl_client_auth": False,
+                            "verify_hostnames": False,
```
reneradoi: Drive-by comment: Hostname verification should not be disabled when using TLS.

.github/workflows/ci.yaml
```diff
@@ -129,6 +129,7 @@ jobs:
           - tests/charms/opensearch_k8s_test_charm
           - tests/charms/opensearch_k8s_2.19.4_test_charm
           - tests/charms/dummy-client-charm
+          - tests/charms/data-integrator
```
akram09, line 132: This is fine for now. But we won't merge this PR until the changes on dataplatformlibs are merged

opensearch_single_kernel/common/client.py
```diff
@@ -539,7 +539,7 @@ def create_user_role(
         resp = self.request(
             "PUT",
             f"{USER_ROLE_ENDPOINT}/{role_name}",
-            payload={**(permissions or {}), **(action_groups or {})},
+            payload={**(permissions or {"index_permissions": []}), **(action_groups or {})},
```
akram09: Why is this ?

opensearch_single_kernel/common/client.py
```diff
@@ -686,7 +686,7 @@ def patch_user(self, user_name: str, patches: list[dict[str, Any]]) -> dict[str,
 
         return resp
 
-    def create_user_role_mapping(self, role: str, mapped_users: list[str]) -> None:
+    def put_role_mapping(self, role: str, users: list[str], backend_roles: list[str]) -> None:
```
akram09, line 696: Update function docstring

opensearch_single_kernel/common/statuses.py
```diff
@@ -389,3 +389,16 @@ class UpgradesStatuses(Enum):
         message="Run `juju trust {charm_app} --scope=cluster`. Needed for in-place refreshes",
         approved_critical_component=True,
     )
+
+
+class LdapStatuses(Enum):
+    """Collection of charm statuses related to LDAP relation."""
+
+    LDAP_DATA_UNAVAILABLE = StatusObject(status="blocked", message="LDAP data is unavailable")
```
akram09: I am not sure if this should be a blocked status or if we should add a status when data from the other charm is not available. Ping @Mehdi-Bendriss

opensearch_single_kernel/common/statuses.py
```diff
@@ -389,3 +389,16 @@ class UpgradesStatuses(Enum):
...
+    LDAPS_NOT_ENABLED = StatusObject(status="blocked", message="LDAPS is not enabled")
```
akram09: This is confusing. If I understand correctly this shows that LDAPS is not enabled from the related ldap charm ? Should change the message if that is the case. Also make sure to follow what Smail did [here](https://github.com/canonical/opensearch-single-kernel-library/pull/128), I meant using short_message, check and action parameters for statuses

opensearch_single_kernel/core/external_clients_relation.py
```diff
@@ -104,3 +111,85 @@ def extra_user_roles(self) -> str:
...
         self.update({"extra-user-roles": roles})
+
+    @property
+    def extra_group_roles(self) -> list[str]:
+        """Get the extra group roles for this relation."""
+        return [
+            role.strip() for role in self.relation_data.get("extra-group-roles", "").split(",")
+        ]
+
+    @extra_group_roles.setter
```
akram09: Why do we have a setter for this ? It should be coming from the data-integrator only

opensearch_single_kernel/core/external_clients_relation.py
```diff
@@ -104,3 +111,85 @@ def extra_user_roles(self) -> str:
...
+    def entity_permissions(self) -> dict[str, list[dict[str, list[str]]]] | None:
+        """Receive, validate and reformat entity permissions into OpenSearch role permissions.
+
+        If request is invalid or absent the None is returned and error is logged.
+        """
+        if not (raw := self.relation_data.get("entity-permissions")):
+            return None
+        try:
+            if (
+                not (parsed := json.loads(raw))
```
akram09: Why don't you use a pydantic model in order to validate this data ?

opensearch_single_kernel/core/external_clients_relation.py
```diff
@@ -5,20 +5,27 @@
...
     Data,
 )
 
+logger = logging.getLogger(__name__)
+
 
 class ExternalOpenSearchClient(RelationState):
     """State collection for a single related external opensearch client."""
 
+    relation: Relation
```
akram09, line 28: Is this a leftover ?

opensearch_single_kernel/core/state.py
```diff
@@ -476,12 +494,11 @@ def external_client_by_relation(self, relation: Relation) -> ExternalOpenSearchC
     @property
     def dashboards_clients(self) -> list[ExternalOpenSearchClient]:
         """Return the dashboard relations out of all."""
-        result = []
-        for external_client in self.external_clients:
-            if (roles := external_client.extra_user_roles) and KIBANA_SERVER_ROLE in roles:
-                # if any(key.name == "opensearch-dashboards" for key in relation.data.keys()):
-                result.append(external_client)
-        return result
+        return [
```
akram09, line 515: Interesting, This is outside of the scope of this PR. But I think we should make this less error prone

opensearch_single_kernel/core/state.py
```diff
@@ -771,25 +788,59 @@ def current_peer_cluster_app(self) -> PeerClusterApp | None:
             ),
         )
 
-    def get_relation_mapped_users(self, role: str) -> list[str]:
-        """Get the list of users mapped to a specific role from config roles_mapping."""
+    @property
+    def roles_mapped_users(self) -> dict[str, list[str]]:
+        """Reconcile the users mapped to roles from roles mapping config."""
+        res = {"own_index": ["*"]}
```
akram09, line 815: We are changing the purpose of this function. What is your reasoning behind this ?

opensearch_single_kernel/core/state.py
```diff
@@ -771,25 +788,59 @@ def current_peer_cluster_app(self) -> PeerClusterApp | None:
...
+        return res
+
+    @property
+    def roles_mapped_roles(self) -> dict[str, list[str]]:
+        """Reconcile the backend roles mapped to roles including client connections."""
+        res = {
+            "manage_snapshots": ["snapshotrestore"],
+            "logstash": ["logstash"],
+            "kibana_user": ["kibanauser"],
+            "all_access": ["admin"],
```
akram09, line 841: Why do we need to have all the default roles in OpenSearch ?

opensearch_single_kernel/core/state.py
```diff
@@ -771,25 +788,59 @@ def current_peer_cluster_app(self) -> PeerClusterApp | None:
...
         except JSONDecodeError:
             logger.error("Bad roles_mapping config value")
-            return []
+            return res
+        for user, role in roles_mapping.items():
+            res.setdefault(role, []).append(user)
+        return res
+
+    @property
+    def roles_mapped_roles(self) -> dict[str, list[str]]:
```
akram09: The function name is confusing

opensearch_single_kernel/core/state.py
```diff
@@ -1272,3 +1323,19 @@ def s3_credentials(self, connection_info: dict[str, str]) -> S3RelDataCredential
...
+        return (
+            self.ldap_requirer.consume_ldap_relation_data(relation=relation)
+            if (relation := self.ldap_relation)
+            else None
+        )
+
+    @property
+    def is_non_main_orchestrator(self) -> bool:
+        """Get whether the current application is a sub-cluster in large deployment."""
+        return (
```
akram09, line 1371: First check we are in a large deployment (Meaning that peer_clusters or peer_clusters_orchestrators have some relatins)

opensearch_single_kernel/events/external_clients.py
```diff
@@ -43,7 +51,10 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
         )
 
         self.framework.observe(
-            self.opensearch_provides.on.index_requested, self._on_index_requested
+            self.opensearch_provides.on.index_requested, self._on_client_requested
```
akram09, line 54: nit: I would prefer to have separate events handling. There maybe some code duplication but It is at least more clear

opensearch_single_kernel/events/external_clients.py
```diff
@@ -65,23 +76,78 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
-        if not self.charm.cluster_manager.opensearch_client.is_node_up() or not event.index:
+        if not self.charm.cluster_manager.opensearch_client.is_node_up():
             event.defer()
-            return
-        if not (external_client := self.charm.state.external_client_by_relation(event.relation)):
-            logger.error("No external client found for relation id %d", event.relation.id)
-            return
+            return False
+
+        if not event.index:
```
akram09, line 148: question: But when we receive an entityRequested event we don't have an index no ?

opensearch_single_kernel/events/external_clients.py
```diff
@@ -133,10 +199,29 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
 
+        external_client.index = external_client.index
+
+        return True
+
+    def _create_client_group(
+        self, event: RelationEvent, external_client: ExternalOpenSearchClient
+    ) -> bool:
+        """Provide the requested group entity for client relation."""
+        if not (entity := external_client.get_requested_entity()):
```
akram09, line 186: I think this should be in get_statuses of Manager. And Why do we show a status that user creation failed ? I think we should point the user to actually specify the right parameters in data-integrator

opensearch_single_kernel/events/external_clients.py
```diff
@@ -133,10 +199,29 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:  # noqa: C901
...
+                dynamic_params={"id": event.relation.id},
+            )
+            event.defer()
+            return False
+
         try:
-            username, pwd = self.charm.external_clients_manager.create_opensearch_users(
-                external_client, event.index, extra_user_roles=event.extra_user_roles
+            self.charm.external_clients_manager.put_client_user(
+                event.relation.id, entity[0], entity[1], external_client.entity_permissions
```
akram09: nit: Create a named tuple for the RequestedEntity

opensearch_single_kernel/events/ldap.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        self.framework.observe(
+            self.charm.state.ldap_certificate_transfer_requires.on.certificate_removed,
+            self._on_ldap_certificate_removed,
+        )
+        self.framework.observe(self.charm.on.config_changed, self._on_config_changed)
+
+    def _on_ldap_ready(self, event: LdapReadyEvent) -> None:
+        """Handle the LDAP integration event."""
+        if not self.charm.workload.can_connect:
```
akram09, line 54: can connect doesn't mean that opensearch is up.

opensearch_single_kernel/managers/config.py
```diff
@@ -428,6 +405,72 @@ def _security_authc_oauth_config(self) -> dict[str, Any]:
...
+        """Get set of LDAP security config options for the Opensearch.
+
+        Intended for authc category in opensearch-security/config.yml config file.
+        """
+        return (
+            {
+                "ldap": {
+                    "http_enabled": True,
+                    "transport_enabled": True,
+                    "order": 1,
```
akram09, line 418: This is a problem, if we give order 1 for all the configs it will cause an issue

opensearch_single_kernel/managers/config.py
```diff
@@ -468,6 +511,44 @@ def _security_authz_static_config() -> dict[str, Any]:
...
+                            "userrolename": self.state.config.get("ldap_user_rolename", ""),
+                            "rolename": self.state.config.get("ldap_role_name_attr", ""),
+                            "rolesearch_enabled": False,
+                            "resolve_nested_roles": False,
+                        },
+                    },
+                },
+            }
+            if (ldap_data := self.state.ldap_data) and ldap_data.ldaps_urls
+            # and self.workload.exists(self.workload.paths.ldap_chain)
```
akram09, line 549: Leftover ?

opensearch_single_kernel/events/ldap.py
```diff
@@ -0,0 +1,147 @@
...
+            return
+
+        if self.charm.upgrades_manager.in_progress:
+            logger.warning(
+                "Changing config during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
+            )
+            event.defer()
+            return
+
+        if not self.charm.cluster_manager.workload.is_service_started():
```
akram09, line 137: I think we should check if the workload is reachable. Since the service might be stuck in a crashloop

opensearch_single_kernel/managers/cluster.py
```diff
@@ -979,3 +980,17 @@ def _add_app_statuses(self, status_list: list[StatusObject]) -> None:  # noqa: C
             and not self.state.application.is_security_index_initialised
         ):
             status_list.append(PeerClusterStatuses.PEER_CLUSTER_NO_DATA_NODE.value)
+
+        self._add_app_ldap_statuses(status_list)
```
akram09: Please add them in a separate ldap manager

opensearch_single_kernel/managers/config.py
```diff
@@ -428,6 +405,72 @@ def _security_authc_oauth_config(self) -> dict[str, Any]:
...
+                            "verify_hostnames": True,
+                            "pemtrustedcas_filepath": self.workload.paths.ldap_chain.as_posix(),
+                            "hosts": [
+                                url.removeprefix("ldaps://") for url in ldap_data.ldaps_urls
+                            ],
+                            "bind_dn": ldap_data.bind_dn,
+                            "password": ldap_data.bind_password,
+                            "userbase": self.state.config.get("ldap_user_base", ""),
+                            "usersearch": self.state.config.get("ldap_user_search", ""),
+                            "username_attribute": "cn",
```
akram09, line 438: Can we make this as well a config param

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,470 @@
...
+@pytest.mark.abort_on_fail
+@pytest.mark.skip_if_deployed
+async def test_deploy_glauth(
+    ops_test_k8s: OpsTest, k8s_model: Model, substrate: Substrate
+) -> None:
+    """Deploys glauth and all required charms coming with glauth and relate them.
+
+    Then it offers the two relations provided by glauth.
+    """
+    # workaround for https://bugs.launchpad.net/snapd/+bug/2127244
```
akram09: It seems the fix for this is released

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,470 @@
...
+        ops_test_k8s,
+        apps=[LDAP_APP_NAME, POSTGRESQL_K8S, TLS_CERTIFICATES_APP_NAME, LDAP_UTILS_APP_NAME],
+        units_statuses={
+            POSTGRESQL_K8S: [EmptyActiveStatus, EmptyBlockedStatus]
+        },  # postgresql can be in Blocked State because of low free space
+    )
+
+    await apply_ldif(ops_test_k8s, k8s_model)
+
+    # On k8s creating an offer on a remote model fails somehow, so we fallback to juju command.
```
akram09: Why does it work in the other integration tests?

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,470 @@
...
+                with attempt:
+                    result = requests.get(
+                        f"https://{ip}:9200/search-index/_search",
+                        headers={"Authorization": authorization},
+                        verify=False,
+                    )
+                    assert result.status_code == 403
+
+
+async def apply_ldif(ops_test_k8s: OpsTest, k8s_model: Model) -> None:
```
akram09: Please move them to top so we understand they are helpers. And have them as private methods

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,470 @@
...
+                *(
+                    f"offer {k8s_model.info.name}.{TLS_CERTIFICATES_APP_NAME}:certificates {CERT_OFFER}".split()
+                )
+            ),
+        )
+
+
+@pytest.mark.abort_on_fail
+@pytest.mark.skip_if_deployed
+async def test_deploy(
```
akram09, line 171: Please add a second flavor for this test such that we test both large deployment and simple deployment

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,470 @@
...
+    await asyncio.gather(
+        model.applications[SEARCH_ADMIN_DATA_INTEGRATOR_NAME].set_config(
+            SEARCH_ADMIN_DATA_INTEGRATOR_CONFIG
+            | {"requested-entities-secret": search_admin_secret}
+        ),
+        model.applications[SEARCH_READONLY_DATA_INTEGRATOR_NAME].set_config(
+            SEARCH_READONLY_DATA_INTEGRATOR_CONFIG
+            | {"requested-entities-secret": search_readonly_secret}
+        ),
+    )
```
akram09: Please add a test for when we add ldap after we have removed to make sure that everything which was created in ldap when connected is removed (meaning we test if the ldap users has access and then re-relate to see if it works again).

opensearch_single_kernel/common/client.py
s-mlnk: API doesn't accept empty payload

opensearch_single_kernel/core/external_clients_relation.py
s-mlnk: This was done to enforce non-nullable type for relation field.

opensearch_single_kernel/core/state.py
s-mlnk: Logic is the same (advanced check was commented out), it's just more readable this way.

opensearch_single_kernel/core/state.py
s-mlnk: Following the specification, the data integrator can request for a role mapping to be created on other OpenSearch roles including default ones such as "kibanauser". Basically it's the list of all available targets for this action.

> image

("search-readonly" is the requirer)

opensearch_single_kernel/core/state.py
s-mlnk: Previously this function was doing a round-trip over config to get mappings for a single specific role provided. If we wanted to reconcile all the mappings (that we should do now with, see https://github.com/canonical/opensearch-single-kernel-library/pull/97#discussion_r3766077793), we would need to call this function for every managed role. Now it aggregates everything at once and we can reuse the result, it's more efficient.
TLDR: from filtering for a single role we now aggregate everything at once.

opensearch_single_kernel/events/external_clients.py
s-mlnk: Wouldn't it be more confusing in this specific case? I mean it's okay to have 2 entry points when we internally react to something, but here we do provide a result (relation change response), and it coming conditionally from the 2 different handlers may be inconvenient.

opensearch_single_kernel/events/external_clients.py
s-mlnk: I've already asked this question earlier 😅. I thought you meant we should use it regardless:

> image
> image

This is the source (you also referenced it):
https://github.com/canonical/opensearch-single-kernel-library/blob/e1222bdb797bcd51ff1e7114ddad1e44b4ff092c/opensearch_single_kernel/lib/charms/data_platform_libs/v0/data_interfaces.py#L5374-L5377

opensearch_single_kernel/core/state.py
s-mlnk: The same check is used in JWT & OAuth, do we need to change that behaviour?

opensearch_single_kernel/events/ldap.py
s-mlnk: Added a check for that

opensearch_single_kernel/managers/config.py
s-mlnk: Do you mean adding parameters to the function? If yes, then why when in all the other places we fill the values directly?

tests/integration/relations/test_ldap.py
s-mlnk: Why separating if we can test both at the same test? Simple deployment test will be literally the same test but some of the checks are cut out. So why then running most of the test twice?

tests/integration/relations/test_ldap.py
s-mlnk: There is already a test for removal (unauthenticated assertion on users after relation departure). Do I need to add re-relation test for all the ldap users tested prior or only for kibana one?

opensearch_single_kernel/managers/config.py
s-mlnk: But we don't have typed charm config, and Mehdi told not to focus on it right now.

tests/integration/relations/test_ldap.py
s-mlnk: We don't follow it and still use small deployments + large deployments testing **without separation** in these tests:
* test_oauth.py
* test_jwt.py
* test_profiles.py

I mean, LD checks will require a SD cluster setup in order to work. So 90% of the test will be shared:
| SD part | LD part |
|--------|--------|
| SD deployment + checks if LDAP works (on SD) | Checks if LDAP works on LD |
| 90% of the current test | 10% of the current test |

So, why do we need to run 90% of this test twice while we can put everything altogether? Otherwise we will run 90% of the test first, and then the same test, but with 10% returned.

All of the measurements are abstract, just to show the point.

opensearch_single_kernel/managers/config.py
```diff
@@ -428,6 +405,73 @@ def _security_authc_oauth_config(self) -> dict[str, Any]:
...
+    def _security_authc_ldap_config(self) -> dict[str, Any]:
+        """Get set of LDAP security config options for the Opensearch.
+
+        Intended for authc category in opensearch-security/config.yml config file.
+        """
+        return (
+            {
+                "ldap": {
+                    "http_enabled": True,
+                    "transport_enabled": True,
```
smethnani: Shouldn't `transport_enabled` be `False`?

opensearch_single_kernel/managers/config.py
```diff
@@ -468,6 +512,44 @@ def _security_authz_static_config() -> dict[str, Any]:
...
+    def _security_authz_ldap_config(self) -> dict[str, Any]:
+        """Get set of LDAP security config options for the Opensearch.
+
+        Intended for authz category in opensearch-security/config.yml config file.
+        """
+        return (
+            {
+                "ldap_roles": {
+                    "http_enabled": True,
+                    "transport_enabled": True,
```
smethnani: Same here shouldn't `transport_enabled` be `False`?

opensearch_single_kernel/managers/config.py
```diff
@@ -468,6 +512,44 @@ def _security_authz_static_config() -> dict[str, Any]:
...
+                                url.removeprefix("ldaps://") for url in ldap_data.ldaps_urls
+                            ],
+                            "bind_dn": ldap_data.bind_dn,
+                            "password": ldap_data.bind_password,
+                            "userbase": self.state.config.get("ldap_user_base", ""),
+                            "usersearch": self.state.config.get("ldap_user_search", ""),
+                            "userrolename": self.state.config.get("ldap_user_rolename", ""),
+                            "rolename": self.state.config.get("ldap_role_name_attr", ""),
+                            "rolesearch_enabled": False,
+                            "resolve_nested_roles": False,
```
smethnani: I'm seeing LDAP being used for the admin user too:
```
juju ssh opensearch/0 'grep "Cannot retrieve roles.*name=admin" /var/log/opensearch/opensearch-*.log'
/var/log/opensearch/opensearch-38sn.log:[2026-09-09T05:44:59,472][ERROR][o.o.s.a.BackendRegistry  ] [opensearch-0.8f0] Cannot retrieve roles for User [name=admin, backend_roles=[admin], requestedTenant=null] from ldap due to OpenSearchSecurityException[OpenSearchSecurityException[No user admin found]]; nested: OpenSearchSecurityException[No user admin found];
```

 We should skip the internal users
```suggestion
                            "resolve_nested_roles": False,
                            "skip_users" : sorted(OPENSEARCH_USERS)
```

opensearch_single_kernel/events/ldap.py
```diff
@@ -0,0 +1,151 @@
...
+            return
```
smethnani: Shouldn't we defer here?

opensearch_single_kernel/events/jwt.py
```diff
@@ -50,10 +49,9 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
 
     def _on_jwt_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle relation creation."""
-        if (
-            deployment_desc := self.charm.state.application.deployment_desc
-        ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
-            # Only the main orchestrator applies JWT config.
+        if self.charm.state.is_non_main_orchestrator:
```
smethnani: Why change the condition here? There can be a window where sub-clusters don't have peer-cluster/peer-cluster-orchestrator relations (for example at deployment or if they simply haven't been related yet).

opensearch_single_kernel/events/ldap.py
```diff
@@ -0,0 +1,151 @@
...
+
+        if not self.charm.cluster_manager.opensearch_client.is_node_up():
+            event.defer()
+            return
+
+        try:
+            self.charm.cluster_manager.apply_security_config(
+                admin_secrets, self.charm.config_manager.SECURITY_CONFIG_YML
+            )
+        except OpenSearchCmdError as e:
```
smethnani: `apply_security_config` already catches and doesn't re-raise `OpenSearchCmdError` so this block will never be reached and the event will not defer. We should instead check if the call returns false (like in the jwt event manager).

opensearch_single_kernel/events/oauth.py
```diff
@@ -67,19 +66,19 @@ def __init__(self, charm: "OpenSearchBaseCharm") -> None:
 
     def _on_oauth_relation_created(self, event: RelationCreatedEvent) -> None:
         """Handler for `relation_created` event."""
-        if (
-            deployment_desc := self.charm.state.application.deployment_desc
-        ) and deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+        if self.charm.state.is_non_main_orchestrator:
```
smethnani: Same here, why change the condition?

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,460 @@
...
+        for authorization in (SEARCH_ADMIN_LDAP_AUTHORIZATION, SEARCH_READONLY_LDAP_AUTHORIZATION):
+            # Wait for LDAP propagation over the cluster
+            for attempt in Retrying(stop=stop_after_delay(600), wait=wait_fixed(10)):
+                with attempt:
+                    result = requests.get(
+                        f"https://{ip}:9200/_plugins/_security/authinfo",
+                        headers={"Authorization": authorization},
+                        verify=False,
+                    )
+                    assert result.status_code == 200
```
smethnani: Please add a check here that the expected roles are in `backend_roles`

opensearch_single_kernel/events/ldap.py
```diff
@@ -0,0 +1,151 @@
...
+        if not self.charm.state.is_non_main_orchestrator:
+            self._update_security_config(event)
+
+    def _on_ldap_certificate_removed(self, event: CertificateRemovedEvent) -> None:
+        """Handle the removal of LDAP certificates."""
+        if not self.charm.workload.can_connect:
+            event.defer()
+            return
+
+        self.charm.workload.unlink(self.charm.workload.paths.ldap_chain)
```
smethnani: I think this will raise an error if the file isn't found
```suggestion
        self.charm.workload.unlink(self.charm.workload.paths.ldap_chain, missing_ok=True)
```

smethnani (commented):
Thanks @s-mlnk! I left some comments

opensearch_single_kernel/managers/config.py
smethnani: Please address this comment. The condition looks incorrect

opensearch_single_kernel/core/state.py
smethnani: I don't think this is correct. We can have non-main-orchestrators in large deployments with no cluster/orchestrator relations there yet.

I don't think we should change behaviour of unrelated features

opensearch_single_kernel/managers/config.py
s-mlnk: Same thing was left in two places, I've only changed one of them and thought I fixed it 🤦‍♂️. Now I fixed it completely.

opensearch_single_kernel/events/ldap.py
s-mlnk: I thought we don't need reloading security config if the whole service will restart here?

opensearch_single_kernel/events/jwt.py
s-mlnk: I just reused this function instead of copy-pasting the same check all over the codebase. About the logic in this function - let's continue conversation in the single thread: https://github.com/canonical/opensearch-single-kernel-library/pull/97#discussion_r3756304920.

opensearch_single_kernel/events/ldap.py
s-mlnk: Nice catch!

opensearch_single_kernel/events/oauth.py
s-mlnk: I just reused this function instead of copy-pasting the same check all over the codebase. About the logic in this function - let's continue conversation in the single thread: https://github.com/canonical/opensearch-single-kernel-library/pull/97#discussion_r3756304920.

tests/integration/relations/test_ldap.py
s-mlnk: I use indirect check for that in `test_ldap_access`. Wouldn't it be redundant?

opensearch_single_kernel/managers/config.py
s-mlnk: Indeed! But what about OAuth & JWT - did we left it accidentally too? @akram09 @Mehdi-Bendriss @skourta

opensearch_single_kernel/events/ldap.py
smethnani: Ah, no, for keystore changes that's true but for any changes to the security index, we need to run securityadmin.sh which doesn't happen on restart.

tests/integration/relations/test_ldap.py
smethnani: I meant more that there's no direct check of backend_roles. So there's no distinction between a permission issue and the backend roles being incorrect. Backend roles should already be assigned at this point in the test

opensearch_single_kernel/events/jwt.py
akram09: I think @smethnani is right, it would be a no-op thing since we don't do much on failover if not started yet.  My comment was coming from two things:
- The fact that on snapshots we are relying on the relations of peer-cluster / peer-cluster-orchestrator 
- And we should not have `not is_non_main_orchestrator` since it does not mean really we are in a main orchestrator.

smethnani: Yes the double negatives are very confusing to read when checking for `not is_non_main_orchestrator`. I would rather the method be `is_main_orchestrator` and only check for specific relations inline if it's ever necessary

opensearch_single_kernel/common/client.py
```diff
@@ -539,7 +543,10 @@ def create_user_role(
         resp = self.request(
             "PUT",
             f"{USER_ROLE_ENDPOINT}/{role_name}",
-            payload={**(permissions or {}), **(action_groups or {})},
+            payload={
+                **(permissions or {"index_permissions": []}),
+                **(action_groups or {}),
```
akram09: What is action_groups, and I think it is not really used

opensearch_single_kernel/core/external_clients_relation.py
akram09: Why do we want that ?

opensearch_single_kernel/core/models.py
```diff
@@ -886,6 +886,47 @@ class SmtpConfig:
...
+                    "allowed_actions": self.privileges,
+                }
+            ]
+        }
+
+
+class ExternalClientEntityPermissions(RootModel[list[ExternalClientEntityPermission]]):
+    """Model class for the entity permissions requested on a client relation."""
+
+    root: list[ExternalClientEntityPermission] = Field(min_length=1, max_length=1)
```
akram09: I don't understand why we have it as a list ? does the Data Integrator provides it as a list ?

opensearch_single_kernel/core/state.py
akram09: Why do we need to have this as default return ?

opensearch_single_kernel/events/ldap.py
```diff
@@ -0,0 +1,148 @@
...
+            ldap_data := self.charm.state.ldap_requirer.consume_ldap_relation_data(
+                relation=event.relation
+            )
+        ):
+            return
+
+        if not ldap_data.ldaps_urls:
+            return
+
+        if not self.charm.workload.exists(self.charm.workload.paths.ldap_chain):
```
akram09: Can you add some logs in all the hook not only here

opensearch_single_kernel/events/ldap.py
akram09: Agree, we should check the service is reachable and if so update the security index. I don't understand why are we talking about restarts here ?

opensearch_single_kernel/managers/config.py
akram09: Thats not what Mehdi was talking about. I am asking if we should have this parameter as a charm config, meaning I can do juju config opensearch ldap-username-attribute="cn". But it is not important, up to @Mehdi-Bendriss to decide

akram09 (commented):
Thank you Serhii, couple of comments. Please don't forget unit tests we really need that !

opensearch_single_kernel/core/external_clients_relation.py
s-mlnk: Just for IDE.

opensearch_single_kernel/events/ldap.py
s-mlnk: Right below there's a check for `is_node_up` which covers your case. But this specific check is meant to skip API reaching if service isn't running at all.

opensearch_single_kernel/managers/config.py
s-mlnk: Ahhh, I confused it with typed charm config, I'm sorry!

opensearch_single_kernel/events/ldap.py
s-mlnk: I confused this part with hot reloading, I'm sorry.

opensearch_single_kernel/common/client.py
s-mlnk: Okay, I removed it.

opensearch_single_kernel/core/models.py
s-mlnk: Yes - [source](https://github.com/canonical/data-platform-libs/blob/02bd35940702e1d095b269f9429278c0f9918bf2/lib/charms/data_platform_libs/v0/data_interfaces.py#L2265).

opensearch_single_kernel/lib/charms/certificate_transfer_interface/v0/certificate_transfer.py
Mehdi-Bendriss: Can you please use the `v1` of this lib?  this will allow us to not only use GLAuth but also Authentik in the future 
https://github.com/canonical/valkey-operator/pull/127/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R14

opensearch_single_kernel/core/state.py
```diff
@@ -1290,3 +1355,19 @@ def s3_credentials(self, connection_info: dict[str, str]) -> S3RelDataCredential
...
+            if (relation := self.ldap_relation)
+            else None
+        )
+
+    @property
+    def is_main_orchestrator(self) -> bool:
+        """Get whether the current application is not a sub-cluster."""
+        return (
+            deployment_desc := self.application.deployment_desc
+        ) is None or deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
```
smethnani: We shouldn't return `True` if the deployment_desc is `None`. That does not mean this is a main orchestrator, it means we can't determine if it's a main orchestrator

opensearch_single_kernel/managers/ldap.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class LdapManager(BaseManager):
+    """OpenSearch LDAP manager class.
+
+    This class is responsible for monitoring the LDAP relation state.
+    """
+
+    def __init__(self, state: ClusterState, workload: BaseWorkload):
+        super().__init__(state, workload, "config_manager")
```
smethnani:
```suggestion
        super().__init__(state, workload, "ldap_manager")
```

opensearch_single_kernel/core/external_clients_relation.py
```diff
@@ -104,3 +115,60 @@ def extra_user_roles(self) -> str:
...
+            return None
+        try:
+            return ExternalClientEntityPermissions.model_validate_json(raw).to_role_permissions()
+        except ValidationError as e:
+            logger.error(
+                "Invalid entity-permissions field in client relation %d: %s",
+                self.relation.id,
+                e,
+            )
+            return None
```
smethnani: Shouldn't we distinguish between invalid permissions and empty permissions? Invalid permissions would still create a user with no permissions

opensearch_single_kernel/core/state.py
```diff
@@ -789,25 +806,68 @@ def current_peer_cluster_app(self) -> PeerClusterApp | None:
...
             if not isinstance(roles_mapping, dict):
                 logger.error("Bad roles_mapping config value")
-                return []
+                return res
         except JSONDecodeError:
             logger.error("Bad roles_mapping config value")
-            return []
+            return res
+        for user, role in roles_mapping.items():
+            res.setdefault(role, []).append(user)
```
smethnani: `"{"alice":["readall"]}"` would be decoded as valid json and put the unit in a hook failure here

opensearch_single_kernel/managers/external_clients.py
```diff
@@ -125,42 +132,34 @@ def get_extra_user_role_permissions(self, extra_user_roles: str, index: str) ->
...
         """
-        try:
-            self.opensearch_client.create_user_role(role_name=user, permissions=permissions)
-        except OpenSearchHttpError as e:
-            raise OpenSearchUserMgmtError(e)
-
-        users = self.state.application.client_users_dict
-
-        if users.get(str(relation_id)):
+        if (users := self.state.application.client_users_dict).get(str(relation_id)):
```
smethnani: Shouldn't we also check if the group is already added by another relation? What if a user relates 2 data integrators with the same group name? The second relation should not overwrite the first.

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,478 @@
...
+            POSTGRESQL_K8S: [EmptyActiveStatus, EmptyBlockedStatus]
+        },  # postgresql can be in Blocked State because of low free space
+    )
+
+    await _apply_ldif(ops_test_k8s, k8s_model)
+
+    if substrate == "vm":
+        await asyncio.gather(
+            k8s_model.create_offer(f"{LDAP_APP_NAME}:ldap", LDAP_OFFER),
+            # GLAuth only speaks certificate-transfer v0; take the CA from the provider that issued
```
smethnani: Is the library not backwards compatible with v0?

smethnani (commented):
Thanks @s-mlnk. I have some questions:

- Is `extra-group-roles` restricted to the charm's fixed list? I tested with:
```
juju config "$ROLE_APP" \
    entity-type=GROUP \
    "requested-entities-secret=$SECRET" \
    extra-group-roles=ml_full_access
juju integrate "$ROLE_APP":opensearch opensearch:opensearch-client
```
I would have expected the role mapping to be created and mapped to the requested group but
`GET /_plugins/_security/api/rolesmapping/ml_full_access` returns 404 not found.

- ~~Is `entity-type=USER` expected to create a relation user? Perhaps I'm missing something, but it's not clear to me why we need to create an opensearch user when `entity-type=GROUP`.~~
- ~~Are permission changes expected to be handled?~~ these were addressed during the daily and not expected to be handled here

There is still not enough testing coverage imo. In addition to the unit tests that Akram requested, can you please add more test coverage including a test which:
- removes and reconnects the ldap and ldap cert relations separately and checks the status and ldap auth and write access in between
- If roles outside of the fixed roles should be supported then tests that show such a role is mapped to the requested group

opensearch_single_kernel/core/external_clients_relation.py
s-mlnk: But we already have a check for empty field:
https://github.com/canonical/opensearch-single-kernel-library/blob/b2d189df660fbba5ae071b23437dcaafd09d56f8/opensearch_single_kernel/core/external_clients_relation.py#L137-L138

tests/integration/relations/test_ldap.py
s-mlnk: Provider should support v1 in this case.

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,478 @@
...
+            POSTGRESQL_K8S,
+            channel="14/stable",
+            trust=True,
+            series="jammy",
+            config={"profile": "testing"},
+        ),
+        k8s_model.deploy(
+            LDAP_APP_NAME,
+            channel="latest/edge",
+            revision=56,
```
akram09: Why revision is fixed ?

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,478 @@
...
+            trust=True,
+        ),
+        k8s_model.deploy(LDAP_UTILS_APP_NAME, channel="latest/edge", trust=True),
+        k8s_model.deploy(
+            TLS_CERTIFICATES_APP_NAME,
+            channel=TLS_STABLE_CHANNEL,
+            trust=True,
```
akram09: I don't think self-signed-certificates requires trust

tests/integration/relations/test_ldap.py
```diff
@@ -0,0 +1,478 @@
...
+        ),
+        k8s_model.deploy(TRAEFIK_CHARM, trust=True),
```
akram09: Please fix the channel here so that we don't get unexpected changes in future

opensearch_single_kernel/managers/external_clients.py
s-mlnk: Done! Although there's important change: I've moved user creation above index creating to avoid index creation on conflicting relations and get rid of "index requested" status flickering.

